Issue #1, opened 2019-02-09
New feature: Block Producer


kstola2:
**Workspace:**
- producer/src/producer/producer.go
- producer/src/producer/producer_test.go
- producer/src/main/main.go

**Objective:**
The purpose of a block producer (_BP_) is to collect data, validate it, bundle it and add it to the ledger. For this issue the focus is on the collection part.

The BP listens for incoming TCP connections and handles them once accepted. A simple echo server is all we need to get started with communication for the BP.

**NOTE:** This code will be running indefinitely on an ACM virtual machine so keep that in mind.

**Testing:**
Test cases that must be passed:

 1. (COMPLETED) If I start up a producer, it should first check if there is an internet connection. If there isn't an internet connection, I shouldn't be able to proceed. Some ways to handle this include a loop that runs infinitely until an internet connection is established or for a certain amount of time before it gives up and exits.
2. (COMPLETED) If someone tries to connect to me, I should accept the connection. It may help to store the connection in a channel.
3. (COMPLETED) If someone is connected to me and sends me a message, I should send it back and that someone should receive a copy of the message they sent me.
4. If I do a keyboard interrupt it should terminate the program gracefully. This means the listener should shut down before the program exits.

**Resources:**
- [Go Testing Package](https://golang.org/pkg/testing/)
- [Go Net Package](https://golang.org/pkg/net/)
- [Go Logging Package](https://golang.org/pkg/log/) (for detailed debug printing)
- [Go Signals](https://gobyexample.com/signals)
- [TCP Server/Client example in Go](https://systembash.com/a-simple-go-tcp-server-and-tcp-client/)

kstola2:
Found this, probably the best way to implement the issue.

https://dev.to/williamhgough/how-to-test-tcpudp-connections-in-go---part-1-3bga

kstola2:
Completed.

PR #17, opened 2019-02-16
Finished up handler test

kstola2:
Issue #1 Task 3

Handler test should run smoothly.


jldistefano:
The test file works for me, though the use of relative import paths concerns me. Because of the way that go's hyperspecific directory tree works, I was only able to run file inside the directory they are contained in.
This can be fixed by replacing any relative import paths with paths from go/src, though it forces you to assume that project_aurum is directly inside of src. An example of this would be instead of `../producer/src` assuming you are in producer, `project_aurum/producer/src` allows this to be run from anywhere, provided project_aurum is directly inside of go/src.

Vineet77:
Yeah, we could go from relative imports to imports from the GO_PATH. It might be better, since then, all people would need to do is go get, to get our modules.

kstola2:
Will need to restructure this branch's directories per PR #19

kstola2:
Made new PR.

PR #19, opened 2019-02-19
Task 1 of Issue 10 Complete

bashbashbash:
I also added in package and imports - these probably need correcting


bashbashbash:
Updated files based on change request - ready for review

bashbashbash:
Ran go test -v on all non main directories and all passed

Issue #11, opened 2019-02-16
Client: Online Status


jldistefano:
**NOTE:** This requires issue #9 to be complete.

**Workspace:**
- client/src/client/client.go
- client/src/client/client_test.go
- client/src/main/main.go

**Objective:**
In order for the user to be able to use the program, the user must be online. 

**Testing:**
1. There's a function in `producer.go` called `CheckConnectivity`. You can use this function whether the client is online or not. You can either create this function again in `client.go` and test it the same way in `client_test.go`. Regardless, in the client's `main.go`, `CheckConnectivity` should be called. If it returns an `err` not equal to `nil`, it should log the problem and exit the program. Otherwise it should just return gracefully. This is pretty much literally the same thing as the first few lines of `main.go` in `producer/`.

**Resources:**
- [Go Testing Package](https://golang.org/pkg/testing/)
- [Go Net Package](https://golang.org/pkg/net/)
- [Go Logging Package](https://golang.org/pkg/log/) (for detailed debug printing)

kstola2:
Completed.

PR #23, opened 2019-02-20
 Issue #11, Added Online Client Status

jldistefano:
Added CheckConnection(), I tried to use absolute import paths instead of relative so test files can be run from anywhere. This will end up making it much easier to write a master test script.

### DIFF
--- a/client/main.go
+++ b/client/main.go
@@ -1,0 +1,5 @@
+package main
+
+func main() {
+
+}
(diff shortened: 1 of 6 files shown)


kstola2:
@jldistefano One of the pull requests is possibly going to be restructuring the project, can you pull from @bashbashbash 's branch and see if your code still runs with it? Or you can do the safer option and just simply copy-paste the function and its test into the client package.

PR #24, opened 2019-02-20
Issue #11, added connection check.

jldistefano:
Added CheckConnection(), uses absolute import paths


kstola2:
I'm gonna approve @bashbashbash 's changes first then I'll take a look at this. Just FYI for the future, you don't need to create a new pull request after every change, you can simply push to your branch and your Pull Request will update with those changes.

PR #32, opened 2019-02-23
Created scripts/ directory and added merge_with_dev script

arosso3:


kstola2:
So just to understand the intent here, this script will update your _personal_ branch with `dev`, but it won't push to `dev` right?

arosso3:
Yes

Issue #36, opened 2019-02-23
Client: Add Quality of Life commands


jldistefano:
**Workspace:**
- client/src/client/client.go
- client/src/client/client_test.go
- client/src/main/main.go

**Objective**:
Now that we have a console user interface up, it would be nice to add some commands that don't necessarily require functionality from the backend.

**Directive**
*Task 1: clear*
Create a function in client.go to clear the console of any current contents, and adds upper separator to terminal window. Essentially invoke the `clear` command on Unix systems and `cls` on Windows. This will likely require an os check.

In main.go, invoke this function when `clear` is input and at program start.
*Task 2: help*
Create a function in client.go to list all the current available functions, and a brief description of each.

In main.go, invoke this function when `help` is input.
*Task 3: moreinfo*
The name of this command can change maybe `source` or `docs`, but it would be nice to have something to send users to the github page for more info. Send user to github page using default browser.

In main.go, invoke this function when `moreinfo` is input.

**Testing:**
1) Run main on linux distribution, mac, and windows. Ensure consistent behavior across each. If you can think of a way to test this in a test case, go right ahead.
2) Run main, ensure proper output. If you can think of a way to test this in a test case, go right ahead.

**Resources:**
- [Go fmt]( https://golang.org/pkg/fmt/#Sprintf )
- [runtime.GOOS os detection](https://golang.org/pkg/runtime/#GOOS)
- [os/exec](https://golang.org/pkg/os/exec/)
- [Go IO]( https://golang.org/pkg/io/ )
- [Go Browser](https://godoc.org/github.com/pkg/browser)

kstola2:
Completed

PR #48, opened 2019-02-27
Issue #36) Added clear, help, moreinfo

jldistefano:
Issue #36, This needs to be tested on mac, windows, and linux, so I tagged some specific reviewers.

This also adds a new required package, that can be found [here](https://godoc.org/github.com/pkg/browser#OpenURL). It should be cloned so the browser directory is go/src/github.com/pkg/browser`. This would likely be added to the requirements eventually, I'm not sure how redistribution and whatnot works so I don't think it would be wise to install it inside of our package.


kstola2:
image

@jldistefano So I'm trying to pull from your branch but it's having merge conflicts I think. Can you double check to make sure your branch is up to date with dev's project structure?

PR #49, opened 2019-02-28
Issue #23) Added Quality of Life Functions

jldistefano:
This is systems dependent, so requires multiple reviews.


bashbashbash:
Tested in Max OS X environment and ran into issue where browser would not open. Updated client.go GoToWeb() function and request testing in other OS environments to ensure nothing is broken

kstola2:
Typically when running stuff in the terminal they will provide just the string link for more information, I don't think I've ever opened my browser from the terminal because of a program. You would probably need code that configures a default browser and stuff.

jldistefano:
The changes work for me, though looking into the source code of that package, it's basically the code I had before. @kstola2 when you get a chance can you see if Zach's changes will work on windows. Alternatively if you think its a bit too weird that a terminal is opening a browser we could just have it print out the github link.

jldistefano:
The branch is updated with the version that prints the link

kstola2:
[This could be interesting in the future for these types of issues.](https://godoc.org/github.com/fatih/color)

jldistefano:
Yeah that might be really nice.

Issue #12, opened 2019-02-16
Master Test Script


jldistefano:
**Workspace:**
- scripts/

**Objective:**
To make testing go smoothly, we should create a script file that goes through each directory and runs any file with test in its name. It should go in a newly made `scripts/` directory.

kstola2:
Completed

PR #50, opened 2019-03-01
Created Master Test Script

arosso3:
#12

### DIFF
--- a/scripts/run_all_tests.sh
+++ b/scripts/run_all_tests.sh
@@ -2,7 +2,3 @@
 
 # Run all go test files in project directory
 go test ../...
-
-# Prompts for user input when done 
-# (this is so that the console stays open so that user can read any error messages)
-read -p "Press Return to Close..."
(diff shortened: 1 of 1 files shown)


arosso3:
Where are you running the script from where the output stays on the screen? When I run it w/o the input line at the end it just closes the console after running.

kstola2:
I'm using the bash terminal in vscode. It also works if I run it in Ubuntu for Windows.

arosso3:
That makes sense. I've just been running it from file explorer.

kstola2:
Try it in a terminal

arosso3:
Unless I missed something in my testing the script should work now if you open it either directly from file explorer or from a terminal with `./run_all_tests.sh`

arosso3:
Tested it on both windows and Linux

jldistefano:
Actually now that I think about it, its kind of strange that its running main files at all.

kstola2:
You could also do something like "if package does not have name main -> run test"

Issue #26, opened 2019-02-22
New Feature: GetKeys


bashbashbash:
**Workspace:**
- producer/src/keys/key.go
- producer/src/keys/key_test.go
- producer/src/keys/.keys.dat.json
- client/src/keys/key.go
- client/src/keys/key_test.go
- client/src/keys/.keys.dat.json

**Objective**:
We need a way to grab `private` and `public` keys for a given user. The `public` key is for verifying and signing the data. The `private` key is for secure access to associated data. The file that will be referenced is *.keys.dat.json*.

**Directive**
Task 1: Create the file `key.go` and `key_test.go`. There will be no testing
Task 2: The function should be called `getKeys`. You will need to open a file as a json object and parse it into appropriate fields. The field parameters should be `private` and `public` and the values will be the keys associated, which will be _hex strings_, see *[big.Int.SetString](https://golang.org/pkg/math/big/#Int.SetString)*. Return the strings separately in the parsed order. If the file does not exist, you will need to *fatal exit*.

```
{
  private : <value>,
  public: <value>
}
```

**Testing:**
1. Create .keys.dat.json if it does not exist
2. Open .keys.dat.json and check for same values returned by getKeys()
3. Assert( .keys.dat.json.public == getKeys().public )
4. Assert( .keys.dat.json.private == getKeys().private )

**Resources:**
- [Go Sprintf](https://golang.org/pkg/fmt/#Sprintf)
- [Go ecdsa](https://golang.org/pkg/crypto/ecdsa/)
- [Go Testing Package](https://golang.org/pkg/testing/)
- [Go JSON Package]( https://golang.org/pkg/encoding/json/ )
- [Go IO]( https://golang.org/pkg/io/ )

jldistefano:
Completed

PR #57, opened 2019-03-02
Issue #26) Added GetKeys

jldistefano:


kstola2:
Hey so a couple things we found out about keys while you were gone:

The functions in ecdsa use a struct called `PrivateKey`, and this struct [needs to be rebuilt in a specific way like the top answer here.](https://stackoverflow.com/questions/21322182/how-to-store-ecdsa-private-key-in-go)

I think angel has pushed to his branch basically the above code, and the idea is to use that to basically use all that to rebuild the private key from a string stored in the file. We're thinking about whether or not to use json format, since we may just need to store the private key and that's it, unless it takes some time to rebuild the public key from the private key. Regardless, I feel like combining both the GenKeys and GetKeys issue is a good idea.

I'll still take a look at your code but coordinating all of this together is difficult in two separate issues.

jldistefano:
Ok, just let me know what you want to do.

kstola2:
@jldistefano can you reach out to angel so he can update you on what we covered?

PR #61, opened 2019-03-04
Tests and function stubs for blockchain file i/o

kstola2:
Issue to follow PR approval

### DIFF
--- a/producer/src/blockchain/blockchain_test.go
+++ b/producer/src/blockchain/blockchain_test.go
@@ -1,0 +1,1 @@
+package blockchain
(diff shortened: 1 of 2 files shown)


producer/src/blockchain/blockchain_test.go
```diff
@@ -0,0 +1,149 @@
...
+		os.Remove(testFile)
+		os.Remove(testDB)
+	}
+	actualBlock, err := GetBlockByHash(block.HashSHA256(expectedBlock.Serialize()), testFile)
+	if err != nil {
+		t.Errorf("Failed to extract block.")
+		os.Remove(testFile)
+		os.Remove(testDB)
+	}
+	if reflect.DeepEqual(expectedBlock.Serialize(), actualBlock) == false {
```
Vineet77: Use bytes.equal to compare byte slices https://golang.org/pkg/bytes/#Equal

PR #70, opened 2019-03-06
Added Command Line Arg Test

jldistefano:


jldistefano:
I've updated the branch

kstola2:
Everything looks fine however I found a package that already does all this lol:

https://godoc.org/github.com/pborman/getopt

It also supports short and long flags (like `-t` vs `--test`). What do you think?

jldistefano:
Honestly I'd rather implement it manually, I could add support for short and long flags as well. I intended to complete this one myself, unless you think it would be a good Friday one.

kstola2:
Okay. For the readcmdlineargs test, I don't understand the situation where that function should return an error though

jldistefano:
If it recieves an empty input line, it should return an error. This wouldn't normally be possible during normal stdin operation, but because it has the ability to simulate input, that case is required.

kstola2:
If it'll never be possible with stdin then that test seems unnecessary and all tests should assume you have at least one arg I think.

jldistefano:
I could change it to only return the queue and assume that there is one argument and thus no errors, but since the function will only be used internally inside of the ProcessCmdAgs function, I don't think there's any harm in keeping it in.

kstola2:
If the function is going to be called inside the other always wouldnt it be better to combine them? Since you're not returning an error from the first one

kstola2:
If the first one just grabs the arguments you could use the getopt package for that part and then issue could be making sure you're pushing processing the arguments and the test could be focused around handling valid args

PR #75, opened 2019-03-07
HashBlock, Deserialize, and Serialize fix 

HarryL5004:
Updated Serialize function (thanks, Vineet) and the tests for data and dataLen. Added HashBlock and Deserialize functions.

### DIFF
--- a/producer/src/block/block.go
+++ b/producer/src/block/block.go
@@ -37,7 +37,7 @@ func (b *Block) Serialize() []byte { // Vineet
 	i := 86
 	for _, s := range b.Data {
 		//for every data entry, put the legth, and then the data
-		binary.LittleEndian.PutUint32(serializedBlock[i:i+4], uint32(len(s)))
+		binary.LittleEndian.PutUint16(serializedBlock[i:i+4], uint16(len(s)))
 		i += 4
 		copy(serializedBlock[i:i+len(s)], s)
 		i += len(s)
--- a/producer/src/block/block.go
+++ b/producer/src/block/block.go
@@ -22,7 +22,7 @@ func (b *Block) Serialize() []byte { // Vineet
 	//NOTE: 32 bit ints are used to hold lengths; unsigned 16 bit int is used for the length of Data
 	bLen := 86 //size of all fixed size fields
 	for _, s := range b.Data {
-		bLen += 4 + len(s) //4 bytes for length plus the length of an element in Data
+		bLen += 2 + len(s) //2 bytes for length plus the length of an element in Data
 	}
 	serializedBlock := make([]byte, bLen)
 
@@ -37,8 +37,8 @@ func (b *Block) Serialize() []byte { // Vineet
 	i := 86
 	for _, s := range b.Data {
 		//for every data entry, put the legth, and then the data
-		binary.LittleEndian.PutUint16(serializedBlock[i:i+4], uint16(len(s)))
-		i += 4
+		binary.LittleEndian.PutUint16(serializedBlock[i:i+2], uint16(len(s)))
+		i += 2
 		copy(serializedBlock[i:i+len(s)], s)
 		i += len(s)
 	}
@@ -109,7 +109,7 @@ func Deserialize(block []byte) Block {
 
 	for i := 0; i < int(dataLen); i++ { // deserialize each individual element in Data
 		elementLen := int(block[index])
-		index += 4
+		index += 2
 		data[i] = block[index : index+elementLen]
 		index += elementLen
 	}
--- a/producer/src/block/block.go
+++ b/producer/src/block/block.go
@@ -13,24 +13,37 @@ type Block struct {
 	PreviousHash   []byte
 	MerkleRootHash []byte
 	Data           [][]byte
+	DataLen        uint16
 }
 
 // Produces a block based on the struct provided
-func (b *Block) Serialize() []byte {
-	// allocates space for the known variables
-	serializedBlock := make([]byte, 20)
+func (b *Block) Serialize() []byte { // Vineet
+	//calculate the total length beforehand, to prevent unneccessary appends
+	//NOTE: 32 bit ints are used to hold lengths; unsigned 16 bit int is used for the length of Data
+	bLen := 4 + 8 + 8 + 32 + 32 + 2 //size of all fixed size fields
+	for _, s := range b.Data {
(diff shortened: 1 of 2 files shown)


producer/src/block/block.go
```diff
@@ -13,24 +13,37 @@ type Block struct {
...
 }
 
 // Produces a block based on the struct provided
-func (b *Block) Serialize() []byte {
-	// allocates space for the known variables
-	serializedBlock := make([]byte, 20)
+func (b *Block) Serialize() []byte { // Vineet
+	//calculate the total length beforehand, to prevent unneccessary appends
+	//NOTE: 32 bit ints are used to hold lengths; unsigned 16 bit int is used for the length of Data
+	bLen := 4 + 8 + 8 + 32 + 32 + 2 //size of all fixed size fields
```
kstola2: Can we make this a static number and just put it straight into the function call to reduce operation cost?

producer/src/block/block.go
```diff
@@ -13,24 +13,37 @@ type Block struct {
 	PreviousHash   []byte
 	MerkleRootHash []byte
 	Data           [][]byte
+	DataLen        uint16
```
kstola2: Lets put `DataLen` before the `Data` field.

producer/src/block/block.go
```diff
@@ -13,24 +13,37 @@ type Block struct {
...
+	for _, s := range b.Data {
+		//for every data entry, put the legth, and then the data
+		binary.LittleEndian.PutUint32(serializedBlock[i:i+4], uint32(len(s)))
+		i += 4
+		copy(serializedBlock[i:i+len(s)], s)
+		i += len(s)
 	}
+
+	// convert datalen into a byte slice in little endian and add to slice
+	binary.LittleEndian.PutUint16(serializedBlock[len(serializedBlock)-2:], b.DataLen)
```
kstola2: serialize the `DataLen` field before the `Data` field.

producer/src/block/block.go
```diff
@@ -13,24 +13,37 @@ type Block struct {
...
+	//NOTE: 32 bit ints are used to hold lengths; unsigned 16 bit int is used for the length of Data
+	bLen := 4 + 8 + 8 + 32 + 32 + 2 //size of all fixed size fields
+	for _, s := range b.Data {
+		bLen += 4 + len(s) //4 bytes for length plus the length of an element in Data
+	}
+	serializedBlock := make([]byte, bLen)
 
 	// convert the known variables to byte slices in little endian and add to slice
 	binary.LittleEndian.PutUint32(serializedBlock[0:4], b.Version)
 	binary.LittleEndian.PutUint64(serializedBlock[4:12], b.Height)
```
kstola2: I am considering removing the `Height` field, because it limits the rate and amount of blocks we can generate. But that'll be a discussion for later, just putting that idea out there so don't worry about it right now.

producer/src/block/block_test.go
```diff
@@ -165,4 +178,3 @@ func TestDeserialize(t *testing.T) {
 		t.Errorf("Blocks do not match")
```
kstola2: This test is failing.

producer/src/block/block.go
```diff
@@ -78,12 +91,40 @@ func getMerkleRoot(l *list.List) []byte {
 // Concatenate all the fields of the block **header**
 // Return the SHA-256 hash of that concatenation
 func HashBlock(b Block) []byte {
-	// TODO
-	return []byte{}
+	blength := 4 + 8 + 8 + 32 + 32 // calculate the total length of the slice
```
kstola2: Let's make this a constant to reduce operation cost

producer/src/block/block.go
HarryL5004: Done

producer/src/block/block.go
HarryL5004: Done

producer/src/block/block.go
HarryL5004: Done

producer/src/block/block_test.go
HarryL5004: I believe this test is failing is because the dataLen field in the expected block is initialized to 0 and not set to the length of Data.
If I include: expected.DataLen = uint16(len(expected.Data)) in the test function, this test will pass.

producer/src/block/block.go
HarryL5004: Done

producer/src/block/block_test.go
kstola2: Ah I overlooked something. We will need to update the `Deserialize` test. Initialize `DataLen` similarly to how the `TestSerialize` does it.

HarryL5004: In TestSerialize, DataLen is set after the test block is initialized. Should I initialize DataLen along with the other members during the initialization of the test block as well?

kstola2: In Go, if you omit initializing a field it just zeros it out. So you can omit initializing `DataLen` when the other fields are initialized and just initialize it after like 
`expected.DataLen = uint16(len(expected.Data))`

HarryL5004:
Sorry if I misunderstood this, so the maximum size of each data piece would be 8192 bytes?
ex: data [][8192]byte

kstola2:
@HarryL5004 So currently the size of the block is a uint32, DataLen is a uint16, and each size of data is a uint32. I'm talking about max capacities here. So theoretically, you can have a block of size 2^32, which 2^16 elements of 2^32 size, if you have each at max values. Which doesnt make sense because 2^32 < 2^16 * 2^32. So we need to change the size of each data piece to a uint16.

kstola2:
@HarryL5004 
image
for this part, make it put a uint16 instead of a uint32

producer/src/block/block.go
```diff
@@ -37,7 +37,7 @@ func (b *Block) Serialize() []byte { // Vineet
 	i := 86
 	for _, s := range b.Data {
 		//for every data entry, put the legth, and then the data
-		binary.LittleEndian.PutUint32(serializedBlock[i:i+4], uint32(len(s)))
+		binary.LittleEndian.PutUint16(serializedBlock[i:i+4], uint16(len(s)))
 		i += 4
 		copy(serializedBlock[i:i+len(s)], s)
```
kstola2: uint16's are 2 bytes, so this needs to changed from 4 to 2, basically everywhere where it says 4

kstola2:
Awesome good work!

Issue #47, opened 2019-02-27
WorkLoop signal handler


kstola2:
**Workspace:**
- producer/src/producer/producer.go
- producer/src/producer/producer_test.go
- producer/src/main/main.go

**Objective:**
The `WorkLoop` function runs indefinitely, but it should gracefully exit and close the listener when an interrupt signal is sent (SIGINT).

**Testing:**
Test cases that must be passed:

1. If I do a keyboard interrupt it should terminate the program gracefully. This means the listener should call `Close()` before the program exits. Also output a `log` message stating that an interrupt signal was detected and the program is terminating.

**Resources:**
- [Go Signals](https://gobyexample.com/signals)

- [Go Testing Package](https://golang.org/pkg/testing/)
- [Go Net Package](https://golang.org/pkg/net/)
- [Go Logging Package](https://golang.org/pkg/log/) (for detailed debug printing)
- [TCP Server/Client example in Go](https://systembash.com/a-simple-go-tcp-server-and-tcp-client/)

kstola2:
Completed.

PR #78, opened 2019-03-08
Issue #47) Added WorkLoop Signal Handler

jldistefano:


kstola2:
I added a carriage return so the `^C` doesn't get printed. Did you intentionally add a newline character at the end of the log statement? Just wondering.

jldistefano:
I didn't notice, but i used Println() with a \n at the end. So I guess it would print two new lines.

kstola2:
Also, I think the log statement should go inside the work loop function, since we're going to be printing out log statements in there as well. The loop might terminate for other reasons, not always an interrupt. Maybe we should pass the logger in as an argument to `WorkLoop`

kstola2:
Made the changes. Let me know what you think

jldistefano:
It works well

kstola2:
I'm working on using it as an argument instead of the current way it's working

kstola2:
This might cause some merge conflicts with dev, but I think it's a lot cleaner. What do you think?

jldistefano:
Runs good for me

Issue #81, opened 2019-03-08
Client: Fix Argument Parsing


jldistefano:
**Workspace:**
- client/src/client/client.go
- client/src/client/client_test.go
- client/src/main/main.go

**Objective**:
The current way of parsing command line arguments uses a queue of strings, a for loop, and a switch statement. While this works perfectly fine, there happens to be a package `getopt` that can do this. So we should switch over to using this package for the sake of simplicity of implementation.

**Directive**
*Task 1: in main()*
For a start we will create a basic option, helpFlag. A boolean flag that will list the usage for all of the command line arguments. This should activate on flags `?`. 
Instead of passing os.Args to ProcessCmdLineArgs, we now need to create option objects and pass those to `ProcessCmdLineArgs()`. This can be done by creating a struct with all of the options. Currently this will only have a boolean `helpFlag` inside of it.
This struct will now be passed to `ProcessCmdLineArgs`

*Task 2: in ProcessCmdLineArgs()*
This will now have to take our new option struct as opposed to the string slice. Instead of a for loop and switch statement, it will now simply go through a gauntlet of if statements based on the elements of the struct. If the helpFlag is true, then the usage for the command line arguments. If this gauntlet of if statements are passed, return nil.

**Testing:**
- Since we cannot test using `getopt` as it is dependent on os.Args which is not set during `go test`, we can simply remove the ProcessCmdArgs test

**Resources:**
- [Getopt](https://godoc.org/github.com/pborman/getopt/v2#pkg-variables)
- [Go IO]( https://golang.org/pkg/io/ )

jldistefano:
Completed

PR #82, opened 2019-03-09
Issue #81) Fixed ProcessCmdLineArgs()

jldistefano:


kstola2:
I'm getting an import error because of hte change of pathing.

Also can you make sure it doesn't modify `producer_test.go` ?

jldistefano:
I made the changes

PR #88, opened 2019-03-09
SQL statements will now execute in tests

kstola2:
@bashbashbash @ramirez915 @arosso3 This PR needs to be approved in order to fix the blockchain test.


kstola2:
@bashbashbash So before I forgot to have the tests call `statement.Exec()` after the insertion, which means before it wouldn't have inserted anything. Now it will, so the tests are fixed.

Issue #66, opened 2019-03-04
New features: Blockchain File I/O


kstola2:
**Expected Date of Completion:** 10-March-2019

**Workspace:**
- `producer/src/blockchain/blockchain.go`
- `producer/src/blockchain/blockchain_test.go`

**Objective:**
In order to add permanence to the blockchain, we will need to store the data locally. File Input/Output is a big issue that encompasses putting blocks into the local copy of the blockchain and storing metadata about it in a database. 

**Phase 1: COMPLETED**
The first function stub `AddBlock` simply wants you to prepend the size of the block as an _unsigned 32 bit integer_ in byte slice form to the block, open a file in `append` mode, write it to the file, then close it. The corresponding test should only fail if you fail to write to the file.

**Phase 2:**
Now we expand the power of `AddBlock`. `AddBlock` should store 4 fields of metadata as a SQL row insertion into the `sqlite3` database provided. Insert a row with the **position** of the block in the file where it's stored, the **hash** of the block header, the **height** from the header, and the **size** from Phase 1.

The corresponding function stubs and tests are written for adding the blocks and three methods of extraction. This will require a lot of sqlite3 learning so I'll be listing resources below.

**Resources:**
- golang sqlite3: [github.com/mattn/go-sqlite3](https://godoc.org/github.com/mattn/go-sqlite3)
- [Tutorial](https://www.thepolyglotdeveloper.com/2017/04/using-sqlite-database-golang-application/)
- [Tutorial](https://astaxie.gitbooks.io/build-web-application-with-golang/en/05.3.html)

kstola2:
Completed

PR #90, opened 2019-03-09
 Phase 2 in AddBlock Function

arosso3:
#66


producer/src/blockchain/blockchain.go
```diff
@@ -33,30 +34,51 @@ func AddBlock(b block.Block, filename string, databaseName string) error { // Ad
...
+	database, err := sql.Open("sqlite3", databaseName)
+	// Checks if the opening was successful
+	if err != nil {
+		return err
+	}
 
-	// statement, _ := database.Prepare("CREATE TABLE IF NOT EXISTS metadata (height INTEGER PRIMARY KEY, position INTEGER, size INTEGER, hash TEXT)")
-	// statement.Exec()
-	// statement, _ = database.Prepare("INSERT INTO")
+	statement, err := database.Prepare("CREATE TABLE IF NOT EXISTS metadata (height INTEGER PRIMARY KEY, position INTEGER, size INTEGER, hash TEXT)")
```
kstola2: I'm not sure we need this called each time, we can assume at this point that the table is created.

producer/src/blockchain/blockchain.go
```diff
@@ -33,30 +34,51 @@ func AddBlock(b block.Block, filename string, databaseName string) error { // Ad
...
+	}
+	_, err = statement.Exec()
+	if err != nil {
+		return err
+	}
+	statement, err = database.Prepare("INSERT INTO metadata (height, position, size, hash) VALUES (?, ?, ?, ?)")
+	if err != nil {
+		return err
+	}
+	_, err = statement.Exec(b.Height, bPosition, bLen, b.PreviousHash)
```
kstola2: The hash should be the Hash of b, not `b.PreviousHash`

producer/src/blockchain/blockchain.go
arosso3: Do you want to get rid of it or just have it run when its empty?

kstola2: I think we can get rid of this particular statement yeah

arosso3: Ok, but then it fails the test because there is no table to add to

kstola2: The table created in the test doesn't persist into the function call?

arosso3: The first test doesn't create a table

kstola2: Ah okay. Let's fix that

kstola2: Just pushed to your branch, let me know if that fixes it

kstola2: Had to push once more to remove the DB at the end of the first test apologies

arosso3:
Updated function

Issue #83, opened 2019-03-09
Client: Debug Mode and Logging


jldistefano:
**Workspace:**
- client/src/client/client.go
- client/src/client/client_test.go
- client/src/main/main.go

**Objective**:
If you've run the main file recently, you've likely noticed that you get a lot of logging messages you probably don't care about right now.
`LOG: 2019/03/07 19:52:14.246300 main.go:28: Connection check passed.`
We will likely only care about these messages when debugging. So we should implement a way to restrict these messages.

**Directive**
This issue requires us to rethink some of the ways that we are currently handling logging, as we need to be able to pass a logger to client.go. This means we have to create a logger object rather than using the standard logger. With this we will have to redo the current logging statements in main.go to use the logging object, not the standard logger.When this logger is created we need to determine where it will write to based on debug flags.
A `-d` flag should indicate that logs should be sent to os.Stdout
A `-f` flag should indicate that logs should be sent to a file, if there is no filename following, it should write to a file `log.txt` otherwise it should use a user defined filename.
If there are no flags regarding debugging, the logs should be sent to a io.Reader object, that will likely never be accessed.

**Testing:**
- Pass all tests on branch `dev`

**Resources:**
- [Go log SetOutput](https://golang.org/pkg/log/#SetOutput)
- [Writing to files](https://gobyexample.com/writing-files)
- [Go IO]( https://golang.org/pkg/io/ )
- Check out how producer handles logging

jldistefano:
Completed

jldistefano:
Completed

PR #93, opened 2019-03-11
Issue #83) Added Debug Mode

jldistefano:
Please test the `-d` flag to see that log messages appear in console, and `-l` should generate a file and directory in `project_aurum/client/logs/logs.txt`. `-ltestfile.txt` should create a file and directory `project_aurum/client/logs/testfile.txt`.

### DIFF
--- a/client/src/main/main.go
+++ b/client/src/main/main.go
@@ -32,8 +32,11 @@ func main() {
 	}
 	// If the log flag is on, it will send the logs to a file in client/logs
 	if getopt.CommandLine.Lookup('l').Count() > 0 {
-		filepath := os.Getenv("GOPATH") + "/src/project_aurum/client/logs"
-		os.Mkdir(filepath, 0777)
+		filepath := "../../logs"
+		err := os.MkdirAll(filepath, 0777)
+		if err != nil {
+			logger.Fatalln(err)
+		}
 		// If no filename is given, logs.txt
 		if *logFile == "" {
 			filepath += "/logs.txt"
@@ -46,7 +49,6 @@ func main() {
 
 		// If there is any error, do not set the logger. Log an error messgae
 		if err != nil {
-			logger.Println(filepath)
 			logger.Fatalln(err)
 		} else {
 			logger = log.New(f, "LOG: ", log.Ldate | log.Lmicroseconds | log.Lshortfile)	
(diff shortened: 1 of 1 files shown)


client/src/main/main.go
```diff
@@ -4,42 +4,73 @@ import (
 	"log"
 	"os"
 	"github.com/pborman/getopt/v2"
+	"bytes"
+	//"fmt"
+	//"strings"
 
-	client "../client"
+	client "project_aurum/client/src/client"
```
kstola2: This import path is causing my file to fail to compile. Is the old version doing the same for you?

jldistefano: It doesn't do it for me, but that's likely because of where my project is installed. That's just an artifact from the branch talking about using absolute import paths rather than relative. But I'll change it back.

jldistefano: The branch is now updated

kstola2:
So debug works great, but log isn't working for some reason. It just quits if I dont give `-l` and argument, and from what I'm assuming it should generate the `logs.txt` file if I do that right?

jldistefano:
When you say it just quits do you mean if you run the command `./main` without a -l it just crashes the program? And the command `./main l` doesn't create a file for you in `project_aurum/client/logs` called `logs.txt`?

kstola2:
image

Also no `logs` directory is created in any of these cases.

jldistefano:
For a start, if you want to specify a filename using the short flag you have to do `-lfilename`, the long ones use `--logfile=filename`. I've updated the branch with a possible fix, which uses the os.MkdirAll command to make the directory, and uses a relative filepath instead of relying on the GOPATH variable. I talk a lot about the relative pathing stuff, but if this fixes it then it is no less broken then the rest of the functioning code we've got right now. I also removed some accidental debugging print statements. Let me know if that works.

client/src/main/main.go
```diff
@@ -4,42 +4,75 @@ import (
...
+	}
+	// Clears the screen before program execution
+	client.ClearScreen()
+
 	// Check to see if there is an internet connection
 	err := client.CheckConnection()
+	//err := error(nil) // This is used for offline testing
 	if err != nil {
-		log.Fatalln("Connectivity check failed.")
+		logger.Fatalln(err)
```
kstola2: Everything looks good, but why was this error statement changed?

jldistefano: I thought it'd be more descriptive to give the actual error instead of just saying that it failed. The error could give more information than I am capable of.

PR #94, opened 2019-03-12
getBlockByHeight and getBlockByPosition functions

ramirez915:
both functions pass their corresponding tests

### DIFF
--- a/producer/src/blockchain/blockchain.go
+++ b/producer/src/blockchain/blockchain.go
@@ -3,9 +3,9 @@ package blockchain
 import (
 	"database/sql"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
-	"log"
 	"os"
 
 	_ "github.com/mattn/go-sqlite3"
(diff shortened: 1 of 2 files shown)


ramirez915:
Just to make sure, just for the AddBlock for now? Or should also for the other functions?

kstola2:
All functions would be best. Also, change the opening file to assume that the file is already created, so no `create` flag.

ramirez915:
Cool, will work on that. Quick thing, there's a log.Fatalln(), should I leave it as is or replace it with the errors.New()?

kstola2:
Replace it with an error

ramirez915:
Alright

producer/src/blockchain/blockchain.go
```diff
@@ -48,27 +49,30 @@ func AddBlock(b block.Block, filename string, databaseName string) error { // Ad
...
 	}
 
 	database, err := sql.Open("sqlite3", databaseName)
 	// Checks if the opening was successful
 	if err != nil {
-		return err
+		return errors.New("Unable to open sqlite3 database")
 	}
 
+	defer file.Close()
```
ramirez915: I think closing the file was forgotten so I added it in

ramirez915:
I have made the changes to the three functions

kstola2:
Looks like those particular functions are working, now we just need `GetBlockByHash` and the final test.

ramirez915:
getBlockByHash is completed and it passes it's corresponding test. I forgot to merge to dev prior to these changes though

PR #96, opened 2019-03-12
Yield functions

kstola2:
**Expected Date of Completion:** 17-March-2019

**Workspace:**
- `producer/src/contract/contract.go`
- `producer/src/contract/contract_test.go`

**Objective:**
Review the way contracts work by reading through the [slides on them](https://github.com/SIGBlockchain/project_aurum/blob/dev/docs/Contracts.pdf).

In this issue we will start building out `Yields` and their corresponding `Unclaimed Yield Pool`. Yields basically denote who has access to a particular chunk of aurum. The UYP is a database table of yields that have yet to be consumed. When you build a contract, you will remove the yields being claimed (consumed) and insert the yields resulting from that transaction.

 The UYP has 4 fields:
- The height of the block that the contract which has the yield is stored in.
- The hash of the contract the yield is stored in
- The index of the yield in the array of yields in that contract
- The rightful holder of the aurum in that yield (ie. the only authorized claimant)
- The value of that yield

Big picture representation:
image

**Testing:**
Tests are located in `contract_test.go`.

**Resources:**
- [Go ecdsa](https://golang.org/pkg/crypto/ecdsa/)
- [Contracts](https://github.com/SIGBlockchain/project_aurum/blob/dev/docs/Contracts.pdf)


bashbashbash:
Various warning received when testing: 
# _/Users/zlabaschin/project_aurum/producer/src/contract [_/Users/zlabaschin/project_aurum/producer/src/contract.test]
./contract_test.go:46:19: undefined: keys.EncodePublicKey
./contract_test.go:51:42: actual.Recipient undefined (type Yield has no field or method Recipient)
./contract_test.go:54:28: actual.Value undefined (type Yield has no field or method Value)
FAIL	_/Users/zlabaschin/project_aurum/producer/src/contract [build failed]

kstola2:
Those need to be implemented by the assignee

Issue #79, opened 2019-03-08
New feature: Client Send


kstola2:
**Expected Date of Completion:** 10-March-2019

**Workspace:**
- `client/src/client/client.go`
- `client/src/client/client_test.go`

**Objective:**
This function will handle sending a stream of bytes to the producer. It is the first issue in a series that begins to connect the producer and the client together.

**Testing:**
The max length of our messages will be 1024, so for messages longer than that a for loop will need to be implemented while sending a stream.

EDIT: The for loop is no longer necessary so long as we limit client messages to be less than 1KB.

**Resources:**
- [Go net package](https://golang.org/pkg/net/)

jldistefano:
Completed

PR #99, opened 2019-03-12
Issue #79) Added SendToProducer()

jldistefano:
Added ln.Close() to tests in order to be able to run multiple tests. Changed import path of producer since it will not actually work with relative import paths, here are a few resources on why relative import paths actually don't work in go, and why we should really start considering conforming to go package standards and structure the project to be contained in an consistent location, probably `go/src/github.com/SIGBlockchain/project_aurum`.
https://scene-si.org/2018/01/25/go-tips-and-tricks-almost-everything-about-imports/
https://stackoverflow.com/questions/38517593/relative-imports-in-go

### DIFF
--- a/client/src/client/client.go
+++ b/client/src/client/client.go
@@ -59,36 +59,13 @@ func SendToProducer(buf []byte, addr string) (int, error) {
 		conn.Close()
 		return 0, err
 	}
-	counter := 0
-	// While size of buf is > 0, send 1024 byte chunks to conn
-	for len(buf) > 1024 {
-		// Send first 1024 bytes of the buffer
-		n, err := conn.Write(buf[:1024])
-		// If there was an error in sending, close connection and return counter and error
-		if err != nil {
-			conn.Close()
-			return counter, err
-		}
-		// Replace buffer with buffer minus first 1024 bytes
-		buf = buf[1024:]
-		// Incriment counter with written bytes
-		counter += n
+	if len(buf) > 1024 {
+		return 0, errors.New("Attempted to send message of size %d, maximum size of 1024 bytes.")
 	}
-	// If there is anything left in the buffer, send it
-	if len(buf) > 0 {
-		// Write remainder of buffer into connection
-		n, err := conn.Write(buf)
-		// If there was an error in sending, close connection and return counter
-		if err != nil {
-			conn.Close()
-			return counter, err
-		}
-		// Incriment counter with written bytes
-		counter += n	
-	}
+	n, err := conn.Write(buf)
 	// Close connection, return counter and no error
 	conn.Close()
-	return counter, nil
+	return len(buf), nil
 }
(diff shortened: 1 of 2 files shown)


kstola2:
On the topic of restructuring the project, I 100% agree and I know we meant to discuss it last Friday but I've made a note to make it a priority for this Friday's meeting. If you've gathered resources and want to make a bottleneck issue describing what needs to be done feel free to do that.

jldistefano:
Yeah, it would be a pretty quick issue. The only thing is that we have to let people know that the `project_aurum` folder would have to be put into a specific place for the imports to work.

kstola2:
Put a plan together or make some suggestions, whatever you feel will work best for everyone on every platform seamlessly.

client/src/client/client_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"testing"
 	"time"
 
-	producer "../../../producer/src/producer"
+	producer "project_aurum/producer/src/producer"
 )
```
kstola2: I know we've been discussing this but I can't run the tests unless the import path is `../../../producer/src/producer`.

client/src/client/client.go
```diff
@@ -53,8 +53,42 @@ func GetUserInput(text *string, reader io.Reader) error {
 // if any conn write call fails, return how many bytes you wrote and an error
 // if everything works out fine, return how many bytes you wrote and a nil error
 func SendToProducer(buf []byte, addr string) (int, error) {
-	// TODO
-	return 0, errors.New("Incomplete function")
+	// Opens a connection, if connection fails, return 0 and error
+	conn, err:= net.Dial("tcp", addr)
+	if err != nil {
+		conn.Close()
+		return 0, err
```
kstola2: For these error returns, it would be useful for future logging debugging statements if they had custom messages describing what was wrong. I'm requestin that if errors are not `nil` that they be `errors.New("error description")`. This is applying to all functions moving forward and we'll retroactively correct this once we begin testing.

client/src/client/client.go
```diff
@@ -53,8 +53,42 @@ func GetUserInput(text *string, reader io.Reader) error {
...
+	}
+	counter := 0
+	// While size of buf is > 0, send 1024 byte chunks to conn
```
kstola2: After we know `conn` is established successfully, I'd ask we use a `defer` statement since we're using Go and that's what it uses for resource cleanups/destructors. So basically you would only need to write `defer conn.Close()` once and it would run upon exiting the function, no matter where it returns.

kstola2:
I just realized, I don't think that the client will ever send anything larger than 1024 bytes, so the loop is unnecessary and actually will cause problems on the producer side, so we can remove it. We will just need to make a note that when the client is sending messages that they can't exceed 1024 bytes.

PR #104, opened 2019-03-15
Issue #12 

bashbashbash:
This script needs to be tested on Windows and Linux. It also needs to test cleanup feature for executable.

#12


bashbashbash:
I can do it as a shell script but Windows script won't run in linux/os 10 environment and vice versa

bashbashbash:
I just realized that we probably can close this issue because all the script file is doing is running go test ./... or go test .\... depending on the OS. So go already has a recursive test script built into it...

kstola2:
There's a way to detect what OS you're using in shell scripts and adapt your script to that: 

[https://stackoverflow.com/questions/394230/how-to-detect-the-os-from-a-bash-script](https://stackoverflow.com/questions/394230/how-to-detect-the-os-from-a-bash-script)

bashbashbash:
I just wanted to doublecheck that the issue objective is to run test files in each directory and subdirectory of the project (meaning run go test -v in every directory with a test file)?

kstola2:
Yeah that's the goal

bashbashbash:
Ok, finally figured it out - one file and the command is simply make, which will run the tests. Needs to be tested in windows and linux environment.

bashbashbash:
See #111 - closing this branch because binary could not be removed

PR #111, opened 2019-03-19
New branch for issue 12 - makefile for running tests

bashbashbash:
#12 - Need testing confirmation from a windows machine and a linux machine


kstola2:
I had to change the import paths, which I believe we're working on, but after that it's working. Can you make this a shell script though and put it in the `scripts` directory like the `merge` script that's in there?

bashbashbash:
Updated per your request - needs testing on windows and linux

kstola2:
**Before running the script now**

Run `go env` and determine where your `GOPATH` is.

The location of the project will need to be in `GOPATH/github/src/SIGBlockchain/project_aurum`.

If you don't have any of those diretories in your `GOPATH`, make them. 

Then run the `masterTest`. **IMPORTANT:** All the contract tests will fail, this is not a bug, as these tests are supposed to fail until they are implemented. All other tests should pass.

I need @bashbashbash and @jldistefano to confirm this is working for them. It is working for me.

kstola2:
I have absolutely zero idea why that error is happening. I see some sort of thread lock... but none of the blockchain tests use thread locks...

kstola2:
@jdistefano https://github.com/facebook/Surround360/issues/3

bashbashbash:
I also have a problem with producer/blockchain:
../producer/src/blockchain/blockchain.go:73:53: undefined: block.HashBlock

I think this is because of the import from block "github.com/SIGBlockchain/project_aurum/producer/src/block"

I tried running go get on it, but still having the same issue

kstola2:
What do your imports look like on blockchain.go?

bashbashbash:
After following directions, everything works as expected

PR #118, opened 2019-03-20
Implemented yields

Vineet77:
I commented out tests not related to yields to allow for the yields test to be executed.

### DIFF
--- a/producer/src/contract/contract_test.go
+++ b/producer/src/contract/contract_test.go
@@ -6,6 +6,7 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"database/sql"
+	"encoding/hex"
 	"log"
 	"os"
 	"reflect"
@@ -13,6 +14,7 @@ import (
 
 	"github.com/SIGBlockchain/project_aurum/producer/src/block"
 	"github.com/SIGBlockchain/project_aurum/producer/src/keys"
+	"github.com/stretchr/testify/assert"
 
 	_ "github.com/mattn/go-sqlite3"
 )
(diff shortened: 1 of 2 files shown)


producer/src/contract/contract_test.go
```diff
@@ -58,24 +75,60 @@ func TestMakeYield(t *testing.T) {
...
 
 /* Tests both serialization and deserialization */
 func TestYieldSerialization(t *testing.T) {
 	testPubKey := generatePubKey()
-	expected := MakeYield(testPubKey, 200000)
+	expected := MakeYield(&testPubKey, 200000)
 	serialized := expected.Serialize()
 	deserialized := DeserializeYield(serialized)
-	if !cmp.Equal(deserialized, expected) {
+	if !reflect.DeepEqual(deserialized, expected) {
```
kstola2: Why was this changed? cmp.Equal is better for struct comparison than reflect

Vineet77: For some reason reflect.DeepEqual works, and I also checked the values of the struct and they were equal, but cmp.Equal still returns the opposite.

Vineet77: Also according to the document cmp.Equal is similar to reflect.DeepEqual.

kstola2: If they are similar then they should both pass

Vineet77: Yeah I think I made a mistake (probably missing a !) initially, because both ways work

kstola2:
Looks good to me.

PR #128, opened 2019-03-23
Claims

Vineet77:


producer/src/contract/contract.go
```diff
@@ -26,8 +26,9 @@ func MakeYield(recipient *ecdsa.PublicKey, value uint64) Yield {
 }
 
 /* Inserts Yield into database */
-func InsertYield(y Yield, database string, blockHeight uint32, contractHash []byte, yieldIndex uint8) error {
+func InsertYield(y Yield, database string, blockHeight uint64, contractHash []byte, yieldIndex uint16) error {
```
kstola2: `yieldIndex` should not be a `uint16` there's no possible way we can support contracts with 65k yields.

producer/src/contract/contract.go
```diff
@@ -65,38 +64,108 @@ the block index containing the contract of the claimed yield,
 the index of the claimed yield in the contract,
 and the public key of the claimant
 */
-type Claim struct{}
+type Claim struct {
+	PreviousContractHash []byte
+	BlockIndex           uint64
+	YieldIndex           uint16
```
kstola2: `YieldIndex uint8`

producer/src/contract/contract.go
```diff
@@ -65,38 +64,108 @@ the block index containing the contract of the claimed yield,
...
+	}
+	defer dbConn.Close()
 
-Prioritize yields that are closest to the value parameter,
-ie. MIN(abs(value - yieldValue0), abs(value - yieldValue1), ... abs(value - yieldValueN))
+	//first query the db for any yeilds equal or above
+	sqlQuery := `SELECT * FROM uy WHERE value >= $1 AND holder = $2 ORDER BY value ASC LIMIT 1;`
+	rows, err2 := dbConn.Query(sqlQuery, value, claimantHexHash)
+	if err2 != nil {
+		log.Printf("Error! Unable to execte query: %v\n", err2)
```
kstola2: Should we return here?

producer/src/contract/contract.go
```diff
@@ -65,38 +64,108 @@ the block index containing the contract of the claimed yield,
...
+		return c, rtnErr
+	}
 
-Case 2:
-If the claimed yield exceeds the value, return
-the claim as usual and a custom error struct with
-the difference as a uint64 field called "Change"
+	//else we need to select a yield that has less value
+	sqlQuery = `SELECT * FROM uy WHERE holder = $1 ORDER BY value DESC LIMIT 1;`
+	rows2, err3 := dbConn.Query(sqlQuery, claimantHexHash)
```
kstola2: you dont need to create multiple `rows` and `errs`, you can simply reassign these with a equal sign without a `:`

producer/src/contract/contract_test.go
```diff
@@ -136,184 +137,230 @@ func TestYieldSerialization(t *testing.T) {
...
-	if testClaim.BlockIndex != 35 {
+	if testClaim.BlockIndex != blockHeight {
 		t.Errorf("Block indeces do not match")
 	}
-	if testClaim.YieldIndex != 1 {
+	if testClaim.YieldIndex != yieldIndex {
 		t.Errorf("Yield indeces do not match")
 	}
-	if !cmp.Equal(testClaim.PublicKey, testPubkey) {
+	if !reflect.DeepEqual(testClaim.PublicKey, pubKey) {
```
kstola2: Does this not work with `cmp`?

producer/src/contract/contract_test.go
```diff
@@ -136,184 +137,230 @@ func TestYieldSerialization(t *testing.T) {
...
-	contractHashFloor := block.HashSHA256([]byte("This should not be claimed"))
-	contractHashCeiling := block.HashSHA256([]byte("This should be claimed"))
-	err := InsertYield(testYieldFloor, "testDB.dat", 35, contractHashFloor, 1)
+func TestMakeClaimMultipleYield(t *testing.T) {
+	db := "testDB.db"
+	setUpDB(db)
+	defer tearDown(db)
+
+	amt := []uint64{uint64(100), uint64(200)}
+	pubKey := []ecdsa.PublicKey{generatePubKey(), generatePubKey()}
```
kstola2: Why are we generating two separate public keys if all these yields should be claimable by the same person?

producer/src/contract/contract_test.go
```diff
@@ -136,184 +137,230 @@ func TestYieldSerialization(t *testing.T) {
...
 	}
-	testClaim, err := MakeClaim("testDB.dat", testPubKey, 10000008)
+
+	testClaim, err := MakeClaim(db, pubKey[1], uint64(50))
 	if err == nil {
-		t.Errorf("Should have an error that resembles change")
+		t.Errorf("No change error struct returned")
+	} else {
+		if change, ok := err.(ChangeError); ok {
+			if change.Change != uint64(150) {
```
kstola2: If we prioritize the `100` yield over the `200` yield then the change should be `50` not `150`.

producer/src/contract/contract_test.go
```diff
@@ -136,184 +137,230 @@ func TestYieldSerialization(t *testing.T) {
...
-		t.Errorf("Failed to insert yield")
-	}
-	err = InsertYield(testYieldFloor, "testDB.dat", 35, contractHashCeiling, 1)
+func TestMakeClaimEmpty(t *testing.T) {
+	db := "testDB.db"
+	setUpDB(db)
+	defer tearDown(db)
+
+	amt := []uint64{uint64(100), uint64(200)}
+	pubKey := []ecdsa.PublicKey{generatePubKey(), generatePubKey(), generatePubKey()}
```
kstola2: multiple public keys?

producer/src/contract/contract.go
Vineet77: There's going to be more than 65k yields in a contract?

kstola2: No we should use a uint8. 65k is way too much.

producer/src/contract/contract_test.go
Vineet77: No, apparently cmp does not support unexported fields

producer/src/contract/contract_test.go
Vineet77: It's a test, MakeClaim should claim the yields corresponding to the public key it was given as a parameter.

producer/src/contract/contract_test.go
Vineet77: No, the yield of 100 corresponds to a different public key, MakeClaim should claim yield corresponding to the public key given (the yield with 200)

producer/src/contract/contract_test.go
Vineet77: Yeah there will be yields corresponding to different public keys in the database.

Vineet77:
I changed yield index to uint8

producer/src/contract/contract.go
Vineet77: You right

producer/src/contract/contract_test.go
kstola2: I don't think there would be an instance where a client has yields corresponding to someone else...

Also what happened to prioritizing yields that are min(abs(yieldval - desiredval), ... ) ?

producer/src/contract/contract_test.go
kstola2: Where's the test showing that `MakeClaim` prioritizes certain yield values over others?

producer/src/contract/contract_test.go
kstola2: There can be in the case of the producer, but that wasn't the original intent behind this test case.

producer/src/contract/contract.go
```diff
@@ -38,65 +39,114 @@ func InsertYield(y Yield, database string, blockHeight uint32, contractHash []by
...
-the index of the claimed yield in the contract,
-and the public key of the claimant
-*/
-type Claim struct{}
-
-/*
-Should scan the Unclaimed Yield Pool for a yield
-
-Prioritize yields that are closest to the value parameter,
-ie. MIN(abs(value - yieldValue0), abs(value - yieldValue1), ... abs(value - yieldValueN))
```
kstola2: Where is this being tested?

producer/src/contract/contract.go
```diff
@@ -38,65 +39,114 @@ func InsertYield(y Yield, database string, blockHeight uint32, contractHash []by
...
+
+	var dbHeight uint64
+	var dbContract string
+	var dbIndex uint8
+	var dbHolder string
+	var dbValue uint64
+
+	//first query the db for any yeilds equal or above
+	sqlQuery := `SELECT * FROM uy WHERE value >= $1 AND holder = $2 ORDER BY value ASC LIMIT 1;`
+	rows, err := dbConn.Query(sqlQuery, value, claimantHexHash)
```
kstola2: We should  make one query to grab all yields, then find the mininimum absolute difference between the claimant desired value and the value of all the yields.

Vineet77: Why not find a yield that can fully pay for the claim?

Vineet77: And then if that fails, find yield that can pay the most.

kstola2: Because the smallest possible absolute minimum difference is zero. The yield selection part should be achievable with a single query. Also the chances of getting exact yield values are less and less likely the more you make contracts. Doing the absolute minimum difference means that if there's a yield that is exact, we will grab it,  otherwise we grab whatever is nearest to it. If your bill is 15, and you have a 10, 10, and a 100, you'd probably use claim the 10 then the other 10 to make your change small. If you had a 15 available the absolute minimum difference would pick that instead with one query.

producer/src/contract/contract_test.go
kstola2: https://godoc.org/github.com/google/go-cmp/cmp#AllowUnexported

producer/src/contract/contract.go
Vineet77: But the first query tries to get yields that are equal to or greater than the desired value. If there are no yields that can fully fund the claim, only then is the second query executed.

kstola2:
Closing this for now.

PR #133, opened 2019-03-30
Recover Metadata feature

HarryL5004:
Updated RecoverBlockchainMetadata function


producer/src/blockchain/blockchain.go
```diff
@@ -258,3 +258,84 @@ func GetBlockByHash(hash []byte, filename string, database string) ([]byte, erro
...
+Moves through the ledger file and extracts blocks, repopulating the metadata file
+Test will check to make sure a ledger populated by AddBlock() will result in the same
+metadata table as would result from here
+Return an error if anything I/O operations fail with detailed message
+*/
+func RecoverBlockchainMetadata(ledgerFilename string, metadataFilename string) error {
+	_, err := os.Stat(metadataFilename)
+	if err != nil {
+		// create database
+		_, err = os.Create("Database.db")
```
kstola2: Use the `metadataFilename` instead.

producer/src/blockchain/blockchain.go
```diff
@@ -258,3 +258,84 @@ func GetBlockByHash(hash []byte, filename string, database string) ([]byte, erro
...
+
+		// open database
+		db, err := sql.Open("sqlite3", "testDatabase.db")
+		if err != nil {
+			return errors.New("Failed to open newly created database")
+		}
+		defer db.Close()
```
kstola2: `defer` statements usually go immediately after a successful call to `Open`. Defers get called in a stack-like order once the function returns.

producer/src/blockchain/blockchain.go
```diff
@@ -258,3 +258,84 @@ func GetBlockByHash(hash []byte, filename string, database string) ([]byte, erro
...
+Return an error if anything I/O operations fail with detailed message
+*/
+func RecoverBlockchainMetadata(ledgerFilename string, metadataFilename string) error {
+	_, err := os.Stat(metadataFilename)
+	if err != nil {
+		// create database
+		_, err = os.Create(metadataFilename)
+
+		// open database
+		db, err := sql.Open("sqlite3", "testDatabase.db")
```
kstola2: `"testDatabase.db"` shouldn't exist, only whatever is passed in as `metadataFilename`

HarryL5004: Should I also change the string for `table` in unit test to `testDatabase.db`? Because otherwise when the unit test calls `GetBlockByHeight(n, testBlockchain.dat, testDatabase.db)`, it will definitely return an error.

image

kstola2: I see what you mean. A mistake on my part when I copy-pasted too liberally. For the "get" functions, pass in `blockchain` and `table` instead of the strings that are currently being passed in. That should fix it.

producer/src/blockchain/blockchain_test.go
```diff
@@ -240,3 +240,81 @@ func TestPhaseTwoMultiple(t *testing.T) {
...
+	}
+	err = AddBlock(block1, blockchain, table)
+	if err != nil {
+		t.Errorf("Failed to add block1.")
+	}
+	err = AddBlock(block2, blockchain, table)
+	if err != nil {
+		t.Errorf("Failed to add block2.")
+	}
+	os.Remove(table)
```
kstola2: For some reason, upon conclusion of the test, we still have `table.db` remaining in our package. The `defer` should be removing it, and this line should also initially remove it. Can you put a check here to see if `table` exists, similar to how you implemented it in the function? And if it still exists, fail the test immediately.

kstola2:
For every file generation call there should be a file removal call; for some reason `table.db` persists after the test is finished running.

producer/src/blockchain/blockchain_test.go
HarryL5004: I put a check there right after` os.Remove(table)` and the test still passed. 

image

HarryL5004: I think `table.db` still exist after the unit test has ended is because there isn't a file removal call for `table.db` after the unit test called `RecoverBlockchainMetadata`

kstola2: Try editing the teardown function to check whether it's properly removing the table.

kstola2: The defer statement should remove it no matter what. You could try duplicating the defer statement after the Recover function call maybe.

HarryL5004: The `table.db` file is still there even after I put `os.Remove(table)` at the very end of the unit test.

kstola2: Check your function to make sure you're closing it properly. The os will not remove files still considered open.

HarryL5004: I figured out the issue. Apparently `os.Create()` creates the file and also opens the file.

Issue #139, opened 2019-04-10
Get youngest block feature


kstola2:
**IMPORTANT:**
There is a branch called `get-youngest-block`. Make a branch based off of this branch to work on. Please do **not** push to this branch directly.

**Desired Date of Completion:** 14-April-2019

**Workspace:**
- producer blockchain package

**Objective:**
When creating the newest block, the producer will require information about the latest block, starting all the way from the genesis block. Important information includes the version, timestamp, and height. I've created a new struct called `BlockHeader` in block which will contain the static-sized information of all Aurum Blocks, called the Header. It might be a good idea to eventually make the `Block` struct composed of a `BlockHeader` and `BlockBody`.

The `blockchain` package in the `get-youngest-block` branch has function stubs and unit tests. Please refrain from changing the unit tests. If it is absolutely necessary, please make a comment first, and then we can both discuss changing them.

**Resources:**
- golang sqlite3: [github.com/mattn/go-sqlite3](https://godoc.org/github.com/mattn/go-sqlite3)
- [sqlite3 Tutorial](https://www.thepolyglotdeveloper.com/2017/04/using-sqlite-database-golang-application/)
- [another sqlite3 Tutorial](https://astaxie.gitbooks.io/build-web-application-with-golang/en/05.3.html)

HarryL5004:
In the unit test, should I include `defer` before `tearDown`? Because otherwise it will remove those files immediately after creating them.
![image](https://user-images.githubusercontent.com/42977312/56054405-564f0f00-5d1c-11e9-86e6-9505ffa2b508.png)

Also, in line 374 of `blockchain_test.go`, should I change `GetYoungestBlock` function to `GetYoungestBlockHeader` so that the unit test compares the header of the youngest block rather than comparing a block and a block header?
![image](https://user-images.githubusercontent.com/42977312/56054863-91057700-5d1d-11e9-9313-9d255f964c26.png)

kstola2:
@HarryL5004 yes sorry I must've missed that.

HarryL5004:
Completed

PR #140, opened 2019-04-12
Get Youngest Block Feature #139

HarryL5004:
Updated GetYoungestBlock and GetYoungestBlockHeader, and made changes to the unit test

### DIFF
--- a/producer/src/blockchain/blockchain_test.go
+++ b/producer/src/blockchain/blockchain_test.go
@@ -326,7 +326,7 @@ func TestGetYoungestBlockAndBlockHeader(t *testing.T) {
 	blockchain := "testBlockchain.dat"
 	table := "testTable.dat"
 	setUp(blockchain, table)
-	tearDown(blockchain, table)
+	defer tearDown(blockchain, table)
 	_, err := GetYoungestBlock(blockchain, table)
 	if err == nil {
 		t.Errorf("Should return error if blockchain is empty")
@@ -371,7 +371,7 @@ func TestGetYoungestBlockAndBlockHeader(t *testing.T) {
 	if err != nil {
 		t.Errorf("Failed to add block")
 	}
-	actualBlock1Header, err := GetYoungestBlock(blockchain, table)
+	actualBlock1Header, err := GetYoungestBlockHeader(blockchain, table)
 	if err != nil {
 		t.Errorf("Error extracting youngest block")
 	}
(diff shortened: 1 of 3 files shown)


HarryL5004:
For some reason there's a conflict in the test results.
image

image

kstola2:
Change the second "blocks do not match" to "block headers do not match" so we can identify which one is failing

HarryL5004:
Is this because I didn't squash the commits into a single commit before pull request?

kstola2:
Print out `actualBlockHeader1` and `blockHeader1` to see if they're different maybe?

kstola2:
No I don't think that's the issue (the commit thing you mentioned)

HarryL5004:
From my computer:
image

kstola2:
This is strange because it's saying the test passes fine...

kstola2:
@HarryL5004 Try remaking the PR with a new branch, with your code in it of course. Make sure you pull from dev because `dev` has been updated sicne this PR was made.

HarryL5004:
The build for my new branch is still failing on travis with the same problem.

kstola2:
@HarryL5004 Yeah that's really weird. For the "Block headers" comparison, try using `reflect.DeepEqual` instead of `cmp.Equal`, see if that changes anything.

kstola2:
What's interesting is when I print out the two block headers in the test, the test fails.

kstola2:
image

HarryL5004:
The timestamps are different..

producer/src/blockchain/blockchain_test.go
```diff
@@ -321,3 +321,61 @@ func TestRecoverBlockchainMetadata(t *testing.T) {
...
+		Timestamp:      time.Now().UnixNano(),
+		PreviousHash:   block.HashSHA256([]byte{'0'}),
+		MerkleRootHash: block.HashSHA256([]byte{'1'}),
+		Data:           [][]byte{block.HashSHA256([]byte("xoxo"))},
+	}
+	block1.DataLen = uint16(len(block1.Data))
+	block1Header := block.BlockHeader{
+		Version:        1,
+		Height:         1,
+		Timestamp:      time.Now().UnixNano(),
```
kstola2: @HarryL5004  I think I figured out the issue. I suspected this might be a timestamp issue. The test is running faster on our machines than on Travis CI, so the timestamps are different between `block1` and `block1header`. Try changing this line to `Timestamp: block1.Timestamp`. That should fix it.

HarryL5004: good news, the build passed

kstola2: Okay, go ahead and make a new PR or update this branch with the appropriate changes. Choice is yours.

Issue #130, opened 2019-03-24
Godocs experiment


kstola2:
I've been looking into `godocs` as a potential way to make a uniform project documentation standard. For this issue, make a branch and follow the godocs style of documentation for the `block.go` file (just that one file). If I understand everything correctly you should be able to launch `localhost` after running `godocs` on a file and see neatly ordered browser documentation.

**== ACTUALLY DOING THIS ==**
Go into whatever folder the source code is in. Run the command `godoc -http=:6060` this will send the results of godoc to your local host, or something. Open a browser and go to the address `http://localhost:6060/pkg/[path to source code folder]`.
ie) `http://localhost:6060/pkg/github.com/SIGBlockchain/project_aurum/producer/src/block`.
The code should be there, if not refer to the stack overflow page.

A good start would be cloning this repo and getting it to work:
[Godocs example repo](https://gist.github.com/cthom06/1089368)
[Godoc doc1](https://blog.golang.org/godoc-documenting-go-code)
[Godoc doc2](https://godoc.org/golang.org/x/tools/cmd/godoc)
[Stack Overflow on how to actually do it](https://stackoverflow.com/questions/24514885/how-do-you-serve-simple-documentation-for-go-programs-using-godoc-as-a-webpage)

kstola2:
Completed

Issue #129, opened 2019-03-24
Technical Documentation for Block


kstola2:
`BLOCK.md` technical documentation. 

In the `BLOCK.md` file located in the `block` package, describe the following in structured paragraphs:
- What the purpose of the `block` package is
- What each function does and why it's important to the overall project
- How each unit tests demonstrates desired behavior
- Possible issues, improvements that can be made
-  Plans for the future if any

Remember, the intended audience of this consists of other programmers and those generally interested in knowing how the code works on a high level. It is meant to provide explanations that are more verbose than in-line code documentation.

Issue #103, opened 2019-03-15
Technical Documentation for Client


jldistefano:
**Workspace:**
- `client/src/client/CLIENT.md`

**Objective:**
Currently the documentation on the project is pretty sparse, consisting of mostly inline comments, its time that we populate the markdown files.

Essentially this file should describe, for fellow programmers, what the code does and how to use it. This shouldn't necessarily be a function by function description, but an overall view.

**Resources:**
Go try running client/src/main/main.go

PR #141, opened 2019-04-12
Updated most source files with godoc documentation notation

jldistefano:
Issues  130


kstola2:
Hey so everything looks good, however this does not cover Issues #103 and #129 because those have to do with the `.md` files. But this does work for #130 so I'll approve it.

jldistefano:
Oh yeah, I forgot about the .md files

PR #142, opened 2019-04-13
Accounts

ramirez915:
I got the accounts tests passing and it seems that I managed to get the database properly updating after the contract has been validated. Only thing omitted was the comparison of structs because I was unable to figure out how to use the cmp.Equal() along with the AllowUnexported(). Also I noticed that the updated database is currently not being tested for.


ramirez915:
Alright cool I'll get on it. I left the prints for now so that you can see that things are being updated properly since there wasn't a test for it at the moment. I forgot what I did that caused the table to remain but I will try to look for the issue.

kstola2:
@ramirez915 After doing some research, I don't think that using `reflect.DeepEqual` will be a big problem. Passing an ignore flag doesn't solve the problem of still needing to compare that field, so we would need to `reflect.DeepEqual` on it anyway after `cmp.Equal`ing the other fields. So I think it's alright to uncomment that comparison block in the test and just use `reflect.DeepEqual` instead.

ramirez915:
I cleaned up the code and substituted the `cmp.Equal` function for the `reflect.DeepEqual` function. The database file seems to be removed after each run.

producer/src/accounts/accounts.go
```diff
@@ -0,0 +1,243 @@
...
+increase value of recipient public key hash account by contract value
+*/
+func (c *Contract) UpdateAccountBalanceTable(table string) {
+	tbl, err := sql.Open("sqlite3", table)
+	if err != nil {
+		fmt.Println("Failed to open sqlite3 table")
+	}
+	defer tbl.Close()
+
+	rows, err := tbl.Query("SELECT public_key_hash , balance, nonce FROM account_balances")
```
kstola2: Can you use a `WHERE public_key_hash = ?` here so that you won't need to loop through each item later?

bashbashbash:
What's the status of this PR?

kstola2:
@bashbashbash the status is you and me are working on it

producer/src/accounts/accounts.go
```diff
@@ -0,0 +1,269 @@
...
+	Version         uint16
+	SenderPubKey    ecdsa.PublicKey
+	SigLen          uint8  // len of the signature
+	Signature       []byte // size varies
+	RecipPubKeyHash []byte // 32 bytes
+	Value           uint64
+	Nonce           uint64
+}
+
+const TableName string = "account_balances"
```
kstola2: What is the purpose of this?

PR #143, opened 2019-04-14
Get Youngest Block Feature

HarryL5004:


HarryL5004:
Should I change reflect DeepEqual back to cmp Equal?

kstola2:
@HarryL5004 yes

Issue #169, opened 2019-05-03
Validate Contract incremental development


kstola2:
branch: `angle2.1`

The Validation test will be a large test, so it will require incremental development. The first test case involves zero value contracts. The other test cases are commented out currently, but we will develop the test as we go.

HarryL5004:
Completed

PR #170, opened 2019-05-03
Validate Contract incremental development

HarryL5004:
Issue #169

### DIFF
--- a/internal/producer/src/accounts/accounts.go
+++ b/internal/producer/src/accounts/accounts.go
@@ -309,6 +309,7 @@ func ValidateContract(c *Contract, table string, authorizedMinters [][]byte) (bo
 		return false, nil
 	}
 
+	// verify the signature in the contract
 	// Serialize the Contract
 	serializedContract := block.HashSHA256(c.Serialize(true))
 
@@ -320,9 +321,29 @@ func ValidateContract(c *Contract, table string, authorizedMinters [][]byte) (bo
 		return false, errors.New("Failed to unmarshal signature")
 	}
 
-	// verify the signature in the contract
+	// if ecdsa.Verify returns false, the signature is invalid
 	if !ecdsa.Verify(c.SenderPubKey, serializedContract, esig.R, esig.S) {
 		return false, nil
+	}
+
+	// check insufficient funds
+	pubKeyStr := hex.EncodeToString(block.HashSHA256(keys.EncodePublicKey(c.SenderPubKey)))
+	// create a query for the row that contains the sender's pkhash in the table
+	sqlQuery := fmt.Sprintf("SELECT balance FROM account_balances WHERE public_key_hash= \"%s\"", pubKeyStr)
+	row, err := db.Query(sqlQuery)
+	if err != nil {
+		return false, errors.New("Failed to create row for query")
+	}
+	defer row.Close()
+
+	if row.Next() {
+		// if row is found, retrieve the sender's balance
+		var balance int
+		row.Scan(&balance)
+		// check if the sender's balance is less than the contract amount
+		if balance < int(c.Value) {
+			return false, nil
+		}
 	}
 
 	return false, errors.New("Failed to validate contract")
--- a/internal/producer/src/accounts/accounts.go
+++ b/internal/producer/src/accounts/accounts.go
@@ -326,10 +326,10 @@ func ValidateContract(c *Contract, table string, authorizedMinters [][]byte) (bo
 		return false, nil
 	}
 
-	// check insufficient funds
+	// check insufficient funds and invalid nonce
 	pubKeyStr := hex.EncodeToString(block.HashSHA256(keys.EncodePublicKey(c.SenderPubKey)))
 	// create a query for the row that contains the sender's pkhash in the table
-	sqlQuery := fmt.Sprintf("SELECT balance FROM account_balances WHERE public_key_hash= \"%s\"", pubKeyStr)
+	sqlQuery := fmt.Sprintf("SELECT balance, nonce FROM account_balances WHERE public_key_hash= \"%s\"", pubKeyStr)
 	row, err := db.Query(sqlQuery)
 	if err != nil {
 		return false, errors.New("Failed to create row for query")
@@ -337,11 +337,15 @@ func ValidateContract(c *Contract, table string, authorizedMinters [][]byte) (bo
 	defer row.Close()
 
 	if row.Next() {
-		// if row is found, retrieve the sender's balance
-		var balance int
-		row.Scan(&balance)
+		// if row is found, retrieve the sender's balance and nonce
+		var balance, nonce int
+		row.Scan(&balance, &nonce)
 		// check if the sender's balance is less than the contract amount
 		if balance < int(c.Value) {
+			return false, nil
+		}
+		// check if the nonce + 1 is not equal to the contract nonce
+		if nonce+1 != int(c.Nonce) {
 			return false, nil
 		}
 	}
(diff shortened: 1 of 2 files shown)


HarryL5004:
Disregard the last commit. I don't think I check for unauthorized minting correctly.

kstola2:
@HarryL5004 what's the status? Should I hold off on the next test?

HarryL5004:
I'm a bit confused in how I should check for false minting contract. Should I check the contract's sender pubkey to see if it matches the authorized minter's pubkey? But then I thought minting contracts have nil sender pubkey

HarryL5004:
@kstola2 I'm a bit confused in how I should check for false minting contract. Should I check the contract's sender pubkey to see if it matches the authorized minter's pubkey? But then I thought minting contracts have nil sender pubkey.

kstola2:
@HarryL5004 Just check if the recipient is in the list of `authorizedMiners` :)

kstola2:
To clarify, the condition for a valid minting contract is that it is in the list of authorized minters AND the `sender` is `nil`.

HarryL5004:
Oh, that makes sense. I didn't understand what the list of authorized minters are for.

HarryL5004:
`ValidateContract` should be good for the next test.

kstola2:
@HarryL5004 I also wanted to know, if there's anything you're confused about conceptually you'd like explained, please don't hesitate to ask. I would be more than happy to explain why things are being done the way they are.

HarryL5004:
Thanks! I guess I am still confused about how `producer` creates the list of authorized minters. Since minting contracts do not have a signature (due to nil sender pubkey), how does `producer` prevents someone from adding himself to the list of authorized minters and giving himself infinite aurum?

`ValidateContract` is good for the next test.

kstola2:
@HarryL5004 so _for now_ the list of authorized minters is controlled by the producer, since there's only going to be one in phase one of Aurum. Basically, we decide who can be a minter or not. The idea is that in a centralized, private system, we call the shots and everybody trusts us. What we want to do in phase two and three is move away from a centralized private system to a decentralized public system. In the future, the list of authorized minters will be elected by the users of Aurum, using Aurum as votes. That way, if someone is being malicious, you can vote them out.

internal/producer/src/accounts/accounts.go
```diff
@@ -277,6 +277,43 @@ func MintAurumUpdateAccountBalanceTable(dbConnection *sql.DB, pkhash []byte, val
...
+	if c.Value == 0 {
+		return false, nil
+	}
+
+	// check for unauthorized minting contracts
+	if c.SenderPubKey == nil {
+		validMinting := false
+		for _, mintersPubKHash := range authorizedMinters {
+			if bytes.Equal(c.RecipPubKeyHash, mintersPubKHash) {
+				validMinting = true
```
kstola2: Would it be cleaner to replace the body of this `if` block with lines 307-312, and then put lines 301-304 after the `for` loop?

internal/producer/src/accounts/accounts.go
```diff
@@ -277,6 +277,43 @@ func MintAurumUpdateAccountBalanceTable(dbConnection *sql.DB, pkhash []byte, val
...
+		}
+		if !validMinting {
+			// unauthorized minting
+			return false, nil
+		}
+
+		// authorized minting
+		err = MintAurumUpdateAccountBalanceTable(db, c.RecipPubKeyHash, c.Value)
+		if err != nil {
+			return false, err
```
kstola2: Can you return a custom error here

kstola2:
Next test is available

internal/producer/src/accounts/accounts.go
HarryL5004: Oh, the reason that I did not return a custom error here is because `MintAurumUpdateAccountBalanceTable` already returns custom errors so I felt that it might be a bit more helpful if I return the func's custom errors instead of making a new custom error.

internal/producer/src/accounts/accounts.go
HarryL5004: Sure thing

internal/producer/src/accounts/accounts.go
kstola2: That's a good point, can you nest the error in the returned error? we can build a custom stack trace that way. the way to do it would be `errors.New("some custom error message: "+err.Error())`, with `err` being the `err` returned from the update function.

HarryL5004: Sure thing

HarryL5004:
I think the test is missing a `defer rows.Close()` statement. It fails to remove the database at the end of the test.

kstola2:
I think you're right, since it's being queried. You can put in the defer statement.

HarryL5004:
ValidateContract is good for the next test.

HarryL5004:
I think the insufficient funds test is missing the arguments for testing.
image

kstola2:
@HarryL5004 updated

HarryL5004:
ValidateContract is good for the next test.

HarryL5004:
ValidateContract is good for the next test.

HarryL5004:
ValidateContract is good for the next test.

kstola2:
aaaand for some reason it's already passing

kstola2:
@HarryL5004 I think this means that the `Exchange...` function is creating the row if it's not there?

HarryL5004:
@kstola2 I doubt it... The `Exchange...` function does not insert any new rows into the database. Should I rewrite the code in `Exchange...` function to see what happens? (The code isn't too well optimized anyways)

kstola2:
Yeah that sounds good. Otherwise this issue is done. The test should fail if that row isn't created, because it's querying the DB for that keyhash that's not inserted...

kstola2:
I added a new test, this one is a contract between the "keyNotInTable" and some new keyhash that is not in the table. I believe it's failing, which is good.

HarryL5004:
I optimized the `Exchange...` function. The test for `ValidateContract` should now properly fails when there's a new account.

HarryL5004:
In the test for the contract between the 'keyNotInTable' account and a newer account, shouldn't the 'keyNotInTable' account nonce be `2` and the newer account's nonce be `1` after the exchange?

kstola2:
No because when we inserted the "keyNotInTable" into the table, it shouldve created that as a new entry in the table, which would instantiate it as zero. Then during this contract, it's sending it's first contract, and subsequently it's next state change, which would increase the nonce by 1.

kstola2:
Also since the "newer" keyhash doesnt exist in the table, it needs to be inserted, so it will be zero, meaning that it is the first contract involving that account.

HarryL5004:
Then shouldn't the nonce for this contract should be `1` instead of `2` because 'keyNotInTable' is sending its first contract?
image

kstola2:
Youre right, I pushed the fix

kstola2:
@Harry I'm considering removing the nonce from the contract, but still having it be updated manually by the functions that control the table. The reason for this is that we have two sets of functions in accounts: those dealing with creating, and those dealing with validating. On the creation side, knowing the nonce requires you to constantly update the nonce and making sure it's corret. On the validation side the most important things to validate would be to just check the value and the signature, and if it's a minting key hash. Do you think it would be a lot of trouble to remove the nonce from the `Contract` struct? 

We would still keep it in the table, and it would just get updated as usual after successful validation.

HarryL5004:
I don't think it would be too much of a trouble to remove the nonce from the `Contract` struct.

kstola2:
@HarryL5004  Okay can you remove it from the struct, but make sure upon successful validation (so only in the functions that are like `Exchange...` or `Insert...`) that the nonce still gets updated? My bad for the lack of foresight, finals are mounting pressure.

HarryL5004:
Sure, no problem.

HarryL5004:
Do you want me to make the changes in the unit tests as well? ('TestMakeContract` and etc. would fail)

kstola2:
Yeah they'll fail otherwise.

kstola2:
Looks good man. Go ahead and slap that on `dev`.

Issue #177, opened 2019-05-12
producer main issue 1/*


kstola2:
branch: `producer main`

package context: `cmd/producer`

Below is a diagram of the states that the producer will go through prior to beginning its infinite loop. It's very important to do all the initialization so that the loop will be as light as possible.
For this issue, the `main.go` file now has a `main_test.go` file, in which we will test `main` being called. Make sure when when gets called, it checks for if there's a wallet, and if there isn't one, make one (#147 ). If there's no ledger, call `panic`. We cannot start the producer without a ledger.

![image](https://user-images.githubusercontent.com/31855040/57576898-f2aa2580-7430-11e9-9fea-e275e3c5ee7f.png)

bashbashbash:
This issue was completed via PR #179

PR #178, opened 2019-05-13
producer main issue 1/* #177

bashbashbash:
Completed - test passed.

Two concerns - this branch is merging into producer-main, not dev. Not sure how we want to handle this.

Since the SetupWallet function is the same as the in the client internal, perhaps we can centralize the function and call it from one place (but this doesn't have to happen now)


bashbashbash:
Cancelled

PR #182, opened 2019-05-15
Dev

kstola2:


kstola2:
@bashbashbash that would probably be best, maybe we can do it tomorrow at the meeting.

jldistefano:
I've run this on my linux machine, the only test that does not pass is `TestBlock_GetHeader`. This is due to an issue regarding timestamps. The timestamps between the expected and actual timestamp differ, but this is normal as they are generated at different times.

kstola2:
@jldistefano I will fix that soon

Issue #173, opened 2019-05-07
Get Account info feature


kstola2:
branch: `query account balance table`

package context: `accounts`

The idea with this is to be able to respond to client account information queries. 

As an additional part of this issue, replace any code in the `accounts` package that this feature overlaps with.

HarryL5004:
For the `AccountInfo_Deserialize` test, ``tt.accInfo`` will always be `nil` regardless of the changes occurred in the `account info deserialize` function.

`857: accInfo: (*AccountInfo)(nil),`
`866: if !reflect.DeepEqual(tt.accInfo, ac) {`

example: https://play.golang.org/p/aGZyFjrJlxA

kstola2:
@HarryL5004 this was the same issue as before with using the cast `nil` pointer, I forgot that this issue had that same problem. I've corrected it and pushed to this branch, so it now should be fine.

HarryL5004:
Completed

PR #184, opened 2019-05-20
Get Account info feature

HarryL5004:
Issue #173

### DIFF
--- a/internal/producer/src/accounts/accounts.go
+++ b/internal/producer/src/accounts/accounts.go
@@ -182,56 +182,31 @@ Add value to recipient's balance
 Increment both nonces by 1
 */
 func ExchangeBetweenAccountsUpdateAccountBalanceTable(dbConnection *sql.DB, senderPKH []byte, recipPKH []byte, value uint64) error {
-	senderQuery := fmt.Sprintf("SELECT balance, nonce FROM account_balances WHERE public_key_hash= \"%s\"", hex.EncodeToString(senderPKH))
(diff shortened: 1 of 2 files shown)


internal/producer/src/accounts/accounts.go
```diff
@@ -392,3 +340,98 @@ func ValidateContract(c *Contract, table string, authorizedMinters [][]byte) (bo
...
+	serializedAccount := make([]byte, 16) // 8 + 8 bytes for balance and stateNonce
+	binary.LittleEndian.PutUint64(serializedAccount[:8], accInfo.balance)
+	binary.LittleEndian.PutUint64(serializedAccount[8:], accInfo.stateNonce)
+	return serializedAccount, nil
+}
+
+func (accInfo *AccountInfo) Deserialize(serializedAccountInfo []byte) error {
+	// accInfo = &AccountInfo{
+	// 	balance:    binary.LittleEndian.Uint64(serializedAccountInfo[:8]),
+	// 	stateNonce: binary.LittleEndian.Uint64(serializedAccountInfo[8:])}
```
kstola2: What is the purpose of these comments?

HarryL5004:
Oh, I thought you meant removing duplicate codes in `account` when you said `replace any code in the accounts package that this feature overlaps with` in the issue. That's why I made the changes to the previous functions...

internal/producer/src/accounts/accounts.go
HarryL5004: Oh, I forgot to remove these comments before committing.. Sorry about that

kstola2:
@HarryL5004 Yeah that's correct what you did I was just double checking the intent behind it. Great work. Just remove those comments in the convo and I'll approve.

Issue #176, opened 2019-05-11
RecoverMetadata support for Account balance Table


kstola2:
branch: `metadata with accounts`

package context: `internal producer`

Firstly, I've moved the `blockchain.RecoverMetadata` function out of the `blockchain` package and into the `producer` package in order to circumvent a circular package dependency issue when modifying the corresponding test. We might as a group consider moving all the `blockchain` package into the `producer` package, since it seems these libraries overlap in purpose.

Now that we can create our genesis block, we need to make sure that upon genesis, both our metadata table and accounts table both are updated appropriately to support further block production and transaction validation. 

The new test for `RecoverMetadata` can be found in the `producer` package, and for now it simply tests making sure that both the metadata table and account balance table are properly initialized with the genesis block and the key hashes within it. A step 2 will be to implement a for loop with many blocks generated, making sure that if `RecoverMetadata` gets called at any point it will have the correct information.

HarryL5004:
Completed

PR #187, opened 2019-05-24
RecoverMetadata support for Account balance Table

HarryL5004:
Issue #176


kstola2:
I have a question: I looked over your code, I'm wondering if the case where the two tables already exist, would calling `Recover...` overwrite them? or would it simply do nothing if it detected them?

HarryL5004:
If the tables already exist, calling `Recover...` will empty the files containing the tables and recover metadata starting from the beginning.

kstola2:
So it will basically overwrite them if they already exist? Cause that would be correct

HarryL5004:
Yea, it should overwrite them if the tables already exist.

kstola2:
@HarryL5004 Update: I put issue #189 up, so that will complicate things less hopefully. You can assume that the []Data field of each block you extract will have serializied contracts, so no more data headers and bodies.

kstola2:
@HarryL5004 Looks good, however because of issue #189 we're going to be replacing the `Data` struct for now (might use it again in a version 2, so that code wasn't all for nothing). This means that these tests and code will need to be tweaked a bit. I'm going to ask that someone finish up that issue so we can knock this one out.

PR #192, opened 2019-05-26
Code using Data struct removed

ramirez915:
I removed the code that contained anything that failed in the remove-Data branch and it seems that every test passes with the modification I had made. I went one by one into the different directories to check what failed and put in the code to make it pass.


ramirez915:
Wrong branch... meant to make the PR to dev

PR #193, opened 2019-05-26
Removed Data struct code

ramirez915:
I replaced the code that had to do with the Data struct, all the tests seem to pass. I checked each directory one by one and ran their tests to see what failed and fixed them.


ramirez915:
Ran all tests running go test -v ./... in project root and all tests pass. Not sure what the issue could be

Issue #202, opened 2019-06-02
Tests for Block comparison function


kstola2:
package context: `internal/block`

We should implement custom equality functions for our structs, so let's start with block. Write a function stub in the above context called `Equals` that takes two `Block` structs as arguments and returns a `bool` determining whether they are equal or not. No need to output an error. Write the corresponding test.

A good test would involve one test case where the blocks are indeed equal, and then a test case for each field where all but one field are equal. Using a custom test-args struct for the test like in the generated go unit tests might be a good idea.

Some interesting resources for this:
https://stackoverflow.com/questions/18926303/iterate-through-the-fields-of-a-struct-in-go
https://golang.org/pkg/reflect/

Once this issue is completed, make an issue to fill out the function body.

HarryL5004:
Should I create a PR to dev for this issue?

kstola2:
Yes or you can just finish the function and then make the PR your choice

HarryL5004:
Completed

PR #207, opened 2019-06-04
Tests for Block comparison function

HarryL5004:
Issue #202
Function stub `Equals` is included.


kstola2:
@HarryL5004 are you going to do the function as well?

HarryL5004:
@kstola2 Yes, I am working on the function right now.

PR #210, opened 2019-06-10
Network producer

kstola2:
Added test for byte channel, communication, and block production, and all are passing


kstola2:
So that file is necessary in order to run the program manually... its contents can be arbitrary. For now we are using 2 hashes that represent Zach's and Harry's wallet addresses. Until we actually launch we can use those, so instead of creating a random file each time we run the program manually, this way we can save time by just reusing a "dummy" hash file. It will be replaced before the air drop.

Issue #216, opened 2019-06-13
`UpdateWallet` function


kstola2:
Given a state nonce and balance, update your wallet fields to the given values.

kstola2:
Completed

Issue #214, opened 2019-06-12
Respond to wallet info request by querying accounts table.


kstola2:
Given a received message that requests wallet info, query the accounts table and send back the balance and state nonce of that wallet address.

kstola2:
I have this working

kstola2:
Completed

PR #217, opened 2019-06-13
Respond to RequestWalletInfo & UpdateWallet

kstola2:
Issues #216  and #214


kstola2:
Okay, I'm probably going to split this into two separate PR's, I'll do that tomorrow.

HarryL5004:
`67: conn, err := ln.Accept()` in `cmd/producer/main.go`
`ln` is `nil` when the package test is running the test for `ResponseToAcc...`, hence the nil pointer error.
Running the test by itself is fine though...

HarryL5004:
` 265: ln, err := net.Listen("tcp", "localhost:9001")`
Changing the port seems to fix the problem..

kstola2:
I think it's cause the goroutines are persisting after each test function concludes. It would make sense that changing the port will make it work.

Issue #212, opened 2019-06-12
ToString function for Contract struct


kstola2:
Create a function in the `accounts` package that uses a `Contract` struct as a receiver and returns a `string`. The string should be newline separated `Contract` values. For any `byte` slices, `hex.Encode` them first.

A useful function for this would be https://golang.org/pkg/fmt/#Sprintf

Make sure that each value is separated by a newline.

bashbashbash:
Complete

PR #218, opened 2019-06-15
`ToString` function for `Contract` struct #212

bashbashbash:
Complete - curious if my test could be stronger


internal/producer/src/accounts/accounts.go
```diff
@@ -454,6 +454,13 @@ func Equals(contract1 Contract, contract2 Contract) bool {
 	return true
 }
 
+// ContractToString takes in a Contract and return a string version
+func ContractToString(c *Contract) string {
+	return fmt.Sprintf("%v\n %v\n %v\n %v\n %v\n %v\n %v\n", c.Version,
```
kstola2: Can we change the format to:

`"Version: %d\nSender Public Key: %s\n..."

bashbashbash:
Replaced with #236

Issue #211, opened 2019-06-12
`ToString` function for `Block` struct


kstola2:
Create a function in the `block` package that uses a `Block` struct as a receiver and returns a `string`. The string should be newline separated `Block` values. For any `byte` slices, `hex.Encode` them first.

A useful function for this would be https://golang.org/pkg/fmt/#Sprintf

Make sure that each value is separated by a newline.

HarryL5004:
Completed

PR #219, opened 2019-06-15
`ToString` function for `Block` struct

HarryL5004:
Issue #211

### DIFF
--- a/internal/producer/src/block/block.go
+++ b/internal/producer/src/block/block.go
@@ -6,6 +6,8 @@ import (
 	"container/list"
 	"crypto/sha256"   // for hashing
 	"encoding/binary" // for converting to uints to byte slices
+	"encoding/hex"
+	"fmt"
 	"reflect"
 )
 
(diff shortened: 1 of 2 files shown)


internal/producer/src/block/block.go
```diff
@@ -206,3 +208,21 @@ func Equals(block1 Block, block2 Block) bool {
...
+	var data string
+	for i, d := range b.Data {
+		if i == len(b.Data)-1 {
+			data += hex.EncodeToString(d)
+		} else {
+			data += hex.EncodeToString(d) + "\n"
+		}
+	}
+
+	blockStr := fmt.Sprintf("%v\n%v\n%v\n", b.Version, b.Height, b.Timestamp)
```
kstola2: Can we change the format to:

`"Version: %d\nHeight: %d\n..."

kstola2:
Looks fine, would you mind screenshotting a sample output of this function?

HarryL5004:
image
This is what the output looks like when the string is printed out.

Issue #213, opened 2019-06-12
Request Wallet info feature sends a message from the client to the producer to update wallet info


kstola2:
Given a flag passed to the client main `-u`, I want to send a message to the producer that requests account info related to the supplied wallet address. Update `aurum_wallet.json` with the response.

**EDIT**:
While working on this issue, I accidentally bricked my motherboard and lost all unsaved changes, so now I need to try and reproduce the steps I took. Here they are:
Steps that need to be completed for this to work:
1. We need to move all _non-main_ functions and tests from the `cmd/producer/main` package into `internal/producer/src/producer` package.
- Note that this may cause errors associated with import cycles; the way I resolved this was looking for any variables/functions in the conflicting files that used the violating import and made sure that that function or variable is using a package definition. An example of this I remember is when moving the `RunServer` function, It typically used `producer.SecretBytes`. However when moving this variable to the `internal/producer/src/producer` package, it should be renamed to simply `SecretBytes`, since they are already in that package. This will require some surgical navigation but I will be available to help. 
- You should also copy the `genesis_hashes.txt` file into the destination package in order to make some tests pass, because if I recall correctly the `ReadGenesisHash` test function deletes this file. Simply make it not delete the file so that the other files will pass.
2. It appears that there are some discrepancies in the functions using the account balance table name among the different packages. In order to guarantee they are all using the same one, we need to change all of them to the same name. A method I chose 
3. After these two steps, run `go test` at the project root to make sure all the tests in Aurum are passing. If they aren't and the above two steps are completed, please let me know and we'll resolve it. IIRC all the tests should pass at this point.
4. I already wrote the function and function body for this issue. It is located [here](https://github.com/SIGBlockchain/project_aurum/blob/requestwalletinfo/internal/client/src/client/client.go) on lines 382-415. A simple copy paste should do.
5. Writing a test for this is going to be tough. Fortunately there already exists a test that mimics this behavior, and that's the response function for this particular request. The `TestResponseToAccountInfoRequest` test located in [dev here](https://github.com/SIGBlockchain/project_aurum/blob/dev/cmd/producer/main/main_test.go) is basically the response version of this request. In order to make this test be feasible, set up a wallet, insert it into a database as an existing account with custom values, then run `go RunServer` to spin up a server. Call the request function, which will receive a `thank you` message followed by the actual response. This response should be parsed to how the producer sends the message in the aforementioned test. Check to make sure it matches the custom values you previously inserted, and it's done.
[...]

PR #225, opened 2019-06-18
Request Wallet info feature

HarryL5004:
Issue #213


cmd/producer/main/main_test.go
```diff
@@ -83,264 +72,3 @@ func TestSuite(t *testing.T) {
...
-		t.Errorf("failed to retrieve wallet address:\n%s", err.Error())
-	}
-	ln, err := net.Listen("tcp", "localhost:10000")
-	if err != nil {
-		t.Errorf("failed to start server:\n%s", err.Error())
-	}
-	byteChan := make(chan []byte)
-	debug := false
-
-	go RunServer(ln, byteChan, debug)
```
kstola2: I believe this is a classic case of two threads racing each other to the end. This is just all the more reason that we should be using docker tests for communication tests. I'm curious, what happens if you put a `time.Sleep()` (https://golang.org/pkg/time/#Sleep) right after `RunServer` ? Set it to like 3 seconds.

HarryL5004: The test passes on my machine, but I'm not sure about Travis.

kstola2: What happens if you run test at project root?

HarryL5004: image
All the tests passed.

kstola2:
It seems to be complaining here:
image

HarryL5004:
Everything seems to be fine after adding a slight delay between the two responses from the server.

cmd/client/main/main.go
```diff
@@ -101,7 +101,33 @@ func main() {
 	}
 
 	if *fl.updateInfo {
-		// TODO: Send a update info message, receive the response and update the wallet balance/nonce
+		fmt.Println("Checking for new wallet info...")
+		accountInfo, err := client.RequestWalletInfo(*fl.producer)
```
kstola2: Can you first check if `*fl.producer` exists? If it doesn't, return a `Fatal` error stating that it's required.

cmd/client/main/main.go
```diff
@@ -101,7 +101,33 @@ func main() {
...
+
+		if currBal != accountInfo.Balance || currNonce != accountInfo.StateNonce {
+			fmt.Println("Updating wallet info...")
+			if err := client.UpdateWallet(accountInfo.Balance, accountInfo.StateNonce); err != nil {
+				lgr.Fatalf("failed to update wallet info: %s", err.Error())
+			}
+		} else {
+			fmt.Println("Wallet is already up to date!")
+		}
+
```
kstola2: Is it necessary to compare the fields? Shouldn't `UpdateWallet` just overwrite whatever is in the wallet? I figure in the case where the fields are the same, overwriting them shouldn't be harmful so long as the private key is the same.

internal/client/src/client/client_test.go
```diff
@@ -507,3 +510,61 @@ func TestUpdateWallet(t *testing.T) {
...
+	_, err := dbc.Exec("CREATE TABLE IF NOT EXISTS account_balances (public_key_hash TEXT, balance INTEGER, nonce INTEGER)")
+	if err != nil {
+		t.Errorf("Failed to create table in database: %s", err)
+	}
+
+	walletAddress, err := GetWalletAddress()
+	if err != nil {
+		t.Errorf("failed to retrieve wallet address:\n%s", err.Error())
+	}
+
```
kstola2: Can you add code to test the case where the wallet address does not exist in the table? It would be before the insertion. You can probably implement a `for` loop and just reuse the code.

cmd/client/main/main.go
HarryL5004: I thought it is always better to avoid unnecessary overwrites, but I guess overwriting some integers here shouldn't be harmful at all. I will remove the comparisons.

kstola2: You have a point. The way I see it is, the producer will always be "correct" in this scenario, so whatever it sends is safe to use for overwriting.

Another way to look at it is we have 2 accessor function calls (`Get...`) that each open and close the wallet, and also 2 comparisons that have the potential to call `UpdateWallet` anyway. Instead we could just call `UpdateWallet`.

internal/client/src/client/client_test.go
kstola2: I only see one call to `RequestWalletInfo`. What I meant was that there should be a case where you call `RequestWalletInfo` and the result from the producer should be a "not found" response.

internal/client/src/client/client.go
```diff
@@ -377,6 +380,41 @@ func GetWalletAddress() ([]byte, error) {
...
+	if _, err := conn.Read(buf); err != nil {
+		return accInfo, errors.New("failed to get thank you message: " + err.Error())
+	}
+	// Should receive message next
+	buf = make([]byte, 1024)
+	nRead, err := conn.Read(buf)
+	if err != nil {
+		return accInfo, errors.New("failed to get response message: " + err.Error())
+	}
+	if buf[8] == 1 {
```
kstola2: This is the case that means "producer did not find the given wallet address in the table"

internal/client/src/client/client.go
```diff
@@ -377,6 +380,41 @@ func GetWalletAddress() ([]byte, error) {
...
+		return accInfo, errors.New("got back failure message from producer")
+	} else if buf[8] == 0 {
```
kstola2: Conversely, this is the case where the producer successfully found the wallet address

internal/client/src/client/client_test.go
kstola2: - [ ] Case where wallet address does not exist should return an `error`
- [ ] Case where wallet address does exist should return `nil`

PR #236, opened 2019-06-20
`ToString` function for `Contract` struct #212

bashbashbash:
Complete

### DIFF
--- a/internal/producer/src/accounts/accounts.go
+++ b/internal/producer/src/accounts/accounts.go
@@ -454,6 +454,13 @@ func Equals(contract1 Contract, contract2 Contract) bool {
 	return true
 }
 
+// ContractToString takes in a Contract and return a string version
+func ContractToString(c *Contract) string {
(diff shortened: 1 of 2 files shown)


internal/producer/src/accounts/accounts.go
```diff
@@ -454,6 +454,13 @@ func Equals(contract1 Contract, contract2 Contract) bool {
 	return true
 }
 
+// ContractToString takes in a Contract and return a string version
+func (c Contract) ToString() string {
+	return fmt.Sprintf("Version: %v\nSenderPubKey: %v\nSigLen: %v\nSignature: %v\nRecipPubKeyHash: %v\nValue: %v\nStateNonce: %v\n",
+						 c.Version, c.SenderPubKey, c.SigLen, hex.EncodeToString(c.Signature), hex.EncodeToString(c.RecipPubKeyHash), 
+						 c.Value, c.StateNonce)
```
kstola2: Can you utf-8 encode the `SenderPubKey` and then hex encode it?

bashbashbash: I read that utf-8 is default, so I tried to convert properly but wasn't sure - check it out

Issue #220, opened 2019-06-17
Remove authorizedMinters


kstola2:
Simply reject any contract where the sender public key is `nil` or if the recipient is the sha-256 hash of the sender public key. Also remove the `authorizedMinters` parameter.

Since phase 1 is centralized, there will be not real need to incentivize minting and thus not required for MVP. This will be a feature in phase 2 instead.

HarryL5004:
Completed

PR #237, opened 2019-06-20
Remove authorizedMinters

HarryL5004:
Issue #220


kstola2:
Do we have a test case that rejects contracts if the recipient is the hash of the sender?

HarryL5004:
`590: senderRecipContract, _ := MakeContract(1, sender, senderPKH, 500, 1)`
`639: {`
`640: name:    "Sender == Recipient",`
`641: c:       senderRecipContract,`
`642: want:    false,`
`643: wantErr: false,`
`644: }`
Yup

PR #242, opened 2019-06-23
Support for contract flag

kstola2:
Contract flag in client works, also validation is done in producer loop.


cmd/client/main/main.go
```diff
@@ -182,10 +182,7 @@ func ContractMessageFromInput(value string, recipient string) ([]byte, error) {
 	}
 
 	// case recipBytes != 32
-	recipBytes, err := hex.DecodeString(recipient)
-	if err != nil {
-		return nil, errors.New("Failed to hex decode recipient")
-	}
```
kstola2: I added this change so that the contracts and update would work

HarryL5004: image
image
Oh, I thought since the test converts `recipientPKH` to a string with `string(..)` we should do the opposite in `ContractMessageFromInput(..)`.

kstola2: I'm not sure about the test but when I did the hex-decode in the function then the actual account info request and contract requests worked.

HarryL5004: Hmm.. when a recipient string is passed into the func, is that string supposed to be hex encoded? Or is it just the result of a string conversion from a byte slice?

kstola2: It is the result of `PrintInfo` in the client's `main.go` function, which prints out a hex-encoded version of the wallet address. So the input into this function should be a hex-encoded string. So we need to hex-decode it to get the right bytes.

kstola2: We should probably add some quality of life improvements in the future to ensure the user is using the correct address. Maybe spellchecking it to make sure the input string is length 64.

HarryL5004: Yea, that would be nice. I guess the `recipient` strings in `TestContractMessageFromInput(...)` need to be changed to be hex encoded as well.

HarryL5004:
Is there anything else we should add/fix for this pull request?

kstola2:
I don't believe so

Issue #227, opened 2019-06-19
Equals function for public keys


kstola2:
Given two `*ecdsa.PublicKey`s, return whether they are equal.

HarryL5004:
Completed

Issue #226, opened 2019-06-19
Equals function for private keys


kstola2:
Given two `*ecdsa.PrivateKey`s, return whether they are equal.

HarryL5004:
Completed

PR #267, opened 2019-07-02
Equals function for public/private keys

HarryL5004:
Issues #226, #227


HarryL5004:
@kstola2 Should I delete the `pkg` directory afterwards?

Issue #269, opened 2019-07-03
pending validation function


kstola2:
See: #268 

context: `validation`

Make a validation function that does the following:

Given a contract, a pending balance, and a pending state nonce:

- reject if the contract's value is zero
- reject if the sender public key is nil or if hashing it matches the recipient
- reject if the signature is invalid
- reject if the pending balance is less than the contract value
- reject if the contract state nonce is not 1 greater than the pending state nonce
- if rejected, return an error
- if valid, return nil for the error, and the new pending balance and state nonce (3 values)
- the new pending balance is `pending balance - contract value`
- the new pending state nonce is `pending state nonce + 1`

HarryL5004:
Completed

Issue #222, opened 2019-06-17
Equals - hash comparison


kstola2:
Given a byte slice and a SHA256 hash, determine whether the hash is the hash of the given byte slice.

Context: `hashing` package.

HarryL5004:
Completed

PR #274, opened 2019-07-03
Pending validation function

HarryL5004:
Issue #269

### DIFF
--- a/internal/producer/src/validation/validation.go
+++ b/internal/producer/src/validation/validation.go
@@ -70,3 +70,55 @@ func ValidateContract(c *contracts.Contract) error {
 
 	return errors.New("Failed to validate contract")
 }
+
+// ValidatePending validates a contract with the given pending balance and pending state nonce
+func ValidatePending(c *contracts.Contract, pBalance *uint64, pNonce *uint64) error {
+	// check for zero value transaction
+	if c.Value == 0 {
+		return errors.New("Invalid contract: zero value transaction")
+	}
+
+	// check for nil sender public key and recip == sha-256 hash of senderPK
+	if c.SenderPubKey == nil || bytes.Equal(c.RecipPubKeyHash, hashing.New(publickey.Encode(c.SenderPubKey))) {
+		return errors.New("Invalid contract: sender cannot be nil nor same as recipient")
+	}
+
+	// verify the signature in the contract
+	// Serialize the Contract
+	copyOfSigLen := c.SigLen
+	c.SigLen = 0
+	serializedContract, err := c.Serialize()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/producer/src/validation/validation.go
```diff
@@ -70,3 +70,55 @@ func ValidateContract(c *contracts.Contract) error {
 
 	return errors.New("Failed to validate contract")
 }
+
+// ValidatePending validates a contract with the given pending balance and pending state nonce
+func ValidatePending(c *contracts.Contract, pBalance *uint64, pNonce *uint64) error {
```
kstola2: Just curious, why are the uint64's passed as pointers?

internal/producer/src/validation/validation.go
```diff
@@ -70,3 +70,55 @@ func ValidateContract(c *contracts.Contract) error {
...
+	// check for zero value transaction
+	if c.Value == 0 {
+		return errors.New("Invalid contract: zero value transaction")
+	}
+
+	// check for nil sender public key and recip == sha-256 hash of senderPK
+	if c.SenderPubKey == nil || bytes.Equal(c.RecipPubKeyHash, hashing.New(publickey.Encode(c.SenderPubKey))) {
```
kstola2: Can we use the new function we made from #222 for this part?

internal/producer/src/validation/validation.go
HarryL5004: I was thinking that rather than returning 0's and an non-nil error for an invalid contract, we can just update the pending balance and state nonce in the func when the validation is successful so that we can prevent any accidental updates to the actual pending balance and state nonce of a sender when the contract validation fails.

internal/producer/src/validation/validation.go
```diff
@@ -79,7 +79,8 @@ func ValidatePending(c *contracts.Contract, pBalance *uint64, pNonce *uint64) er
 	}
 
 	// check for nil sender public key and recip == sha-256 hash of senderPK
-	if c.SenderPubKey == nil || bytes.Equal(c.RecipPubKeyHash, hashing.New(publickey.Encode(c.SenderPubKey))) {
+	recipPKhash := hashing.SHA256Hash{SecureHash: c.RecipPubKeyHash}
```
kstola2: Just FYI I think you can do literals when making structs, so basically these would both be valid:
`recipPKhash := hashing.SHA256Hash{SecureHash: c.RecipPubKeyHash}`
`recipPKhash := hashing.SHA256Hash{c.RecipPubKeyHash}`

If it's multiple fields so long as they're in order it works, but it's just a neat work around I didn't know if you were aware of.

Issue #271, opened 2019-07-03
Add to map function


kstola2:
See: #268 

Add some functionality to the pending pool map struct from issue #270 .

It should be able to return whether a key exists in the map (example in here: https://blog.golang.org/go-maps-in-action)

It should have an add function, that addresses two cases:
- If the contract's sender public key hash is not in the map, pass it to our regular `validateContract` function. If it's invalid, return an error. If it is valid, it'll get the table balance and table nonce, create a pending balance/pending state nonce struct with the pending balance being the difference between the table balance and the valid contract value and pending nonce being equal to the contract nonce, add the hash and the struct to the map, then return `nil`.
- If the contract's sender public key hash _is_ in the map, pass it to the #269 validation function. If that returns an error, return an error. If it returns nil, update the map's struct for that public key hash with the two values returned from #269. Then return `nil`.

HarryL5004:
Completed

Issue #270, opened 2019-07-03
pending pool package


kstola2:
See: #268 

Make a new package pertaining to pending pool.

Make a struct that contains a pending balance and pending state nonce. 

Make a struct that contains a map who's key is a wallet address and who's value is of the type of the above struct.

I don't think there need to be any tests for this. A couple of `New` constructor functions might be nice.

Help with maps: https://blog.golang.org/go-maps-in-action

HarryL5004:
I don't think we can use byte slice as a key for maps in golang, but we can encode the wallet address to string and use that as a key instead.

> ...map keys may be of any type that is comparable...Notably absent from the list are slices, maps, and functions; these types cannot be compared using ==, and may not be used as map keys.
-From https://blog.golang.org/go-maps-in-action

kstola2:
How about we hex encode to string?

HarryL5004:
Completed

Issue #250, opened 2019-06-24
Change accounts package I/O function file parameter


kstola2:
Change I/O related functions to use file pointers/database connections instead of opening and closing excessively in the `accounts` package

https://www.vividcortex.com/blog/2015/09/22/common-pitfalls-go/

Example: [This](https://github.com/SIGBlockchain/project_aurum/blob/dev/internal/producer/src/accountstable/accountstable.go#L110) should instead be a parameter that is passed in

- [x] All `accountstable` package functions use database connections passed in
- [x] All `accountstable` tests pass
- [x] All usages of modified `accountstable` functions refactored

HarryL5004:
Completed

PR #276, opened 2019-07-07
Pending pool package

HarryL5004:
Issue #270, #271

### DIFF
--- a/internal/pendingpool/pendingpool.go
+++ b/internal/pendingpool/pendingpool.go
@@ -1,0 +1,85 @@
+package pendingpool
+
+import (
+	"database/sql"
+	"encoding/hex"
+	"errors"
+
+	"github.com/SIGBlockchain/project_aurum/internal/constants"
+	"github.com/SIGBlockchain/project_aurum/internal/producer/src/contracts"
+	"github.com/SIGBlockchain/project_aurum/internal/producer/src/hashing"
+	"github.com/SIGBlockchain/project_aurum/internal/producer/src/validation"
+	"github.com/SIGBlockchain/project_aurum/internal/publickey"
+)
+
+//PendingData contains pending balance and pending nonce
+type PendingData struct {
+	PendingBal   uint64
+	PendingNonce uint64
+}
+
+//PendingMap contains a map that maps a hex encoded string of a wallet address to a pointer of PendingData
+type PendingMap struct {
+	Sender map[string]*PendingData
+}
+
+//NewPendingData returns an instance of pendingData given pending balance and pending nonce
+func NewPendingData(pendingBal uint64, pendingNonce uint64) PendingData {
+	return PendingData{pendingBal, pendingNonce}
+}
+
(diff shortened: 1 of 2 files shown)


internal/pendingpool/pendingpool.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+//NewPendingData returns an instance of pendingData given pending balance and pending nonce
+func NewPendingData(pendingBal uint64, pendingNonce uint64) PendingData {
+	return PendingData{pendingBal, pendingNonce}
+}
+
+//NewPendingMap returns an instance of pendingMap given a wallet address and an instance of pendingData
+func NewPendingMap(walletAddress []byte, data PendingData) PendingMap {
+	encodedWalletAddr := hex.EncodeToString(walletAddress)
```
kstola2: Can this map be instantiated without passing in anything? Or I guess I don't see why it needs to be instantiated with at least one map element.

internal/pendingpool/pendingpool.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	senderPD, inMap := m.Sender[senderPKStr]
+	if !inMap { // if the key is not in the map
+		err := validation.ValidateContract(c)
+		if err != nil {
+			return errors.New("Failed to validate contract: " + err.Error())
+		}
+
+		accDB, err := sql.Open("sqlite3", constants.AccountsTable)
```
kstola2: Per issue #250 , we will no longer be opening and closing database connections within functions because it's needlessly resource intensive to be opening and closing them for short periods of time. So we can just assume that a database connection is passed into this function.

internal/pendingpool/pendingpool.go
```diff
@@ -0,0 +1,85 @@
...
+		if err != nil {
+			return errors.New("Failed to open accounts table: " + err.Error())
+		}
+		defer accDB.Close()
+
+		row, err := accDB.Query("SELECT balance FROM account_balances WHERE public_key_hash = \"" + senderPKStr + "\"")
```
kstola2: Don't we have a function that retrieves a balance in the `accountstable` package?

internal/pendingpool/pendingpool.go
```diff
@@ -0,0 +1,85 @@
...
+	return PendingMap{m}
+}
+
+//Add returns an error if the process of validating the given contract has failed.
+//Otherwise, Add either inserts the sender's PKhash and the PendingData struct into the map,
+//or updates the pending balance and pending nonce for that sender's PKhash in the map
+func (m *PendingMap) Add(c *contracts.Contract) error {
+	senderPKHash := hashing.New(publickey.Encode(c.SenderPubKey))
+	senderPKStr := hex.EncodeToString(senderPKHash) // hex encoded sender PKhash string for the key
+	if m.Sender == nil {
```
kstola2: Isn't this redundant if the map is already constructed in the `NewPendingMap` function?

Issue #285, opened 2019-07-21
Post ip:port combo to blockpub


kstola2:
Using a GET request, post the producer's IP and port combination to the peer discovery page.

kstola2:
Completed

PR #289, opened 2019-07-27
Get request blockpub add peer

kstola2:
Issue #285

### DIFF
--- a/internal/endpoints/endpoints.go
+++ b/internal/endpoints/endpoints.go
@@ -3,4 +3,5 @@ package endpoints
 const (
 	AccountInfo = "/accountinfo"
 	Contract    = "/contract"
+	AddPeer     = "http://blockchain.acmapp.tech/peer/add"
 )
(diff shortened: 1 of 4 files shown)


kstola2:
> 
> 
> Looks good, but if we are posting the producer's IP and port to the web page you showed us on Monday, shouldn't we use a POST request rather than a GET request?



> 
> 
> Looks good, but if we are posting the producer's IP and port to the web page you showed us on Monday, shouldn't we use a POST request rather than a GET request?

That's a great question. So while GET requests typically deliver some parameter in order to get a response back with some information, it doesn't necessarily always need to restricted to this type of service. If I only want to post say 1 or 2 parameters, like a login, we can do a GET request with those. 

Incidentally, `blockpub` is configured to only accept GET requests on the `peer/add` endpoint:

https://github.com/SIGBlockchain/blockpub/blob/master/routes/web.php#L19

PR #290, opened 2019-07-27
Change blockchain package I/O function file parameter 

HarryL5004:
Issue 249

### DIFF
--- a/cmd/handlers/main.go
+++ b/cmd/handlers/main.go
@@ -92,8 +92,8 @@ func main() {
 	if err != nil {
 		log.Fatalf("Failed to get youngestBlockHeader")
 	}
-	if errClosing := ledgerFile.Close(); errClosing != nil {
-		log.Fatalf("Failed to close ledger file")
+	if err := ledgerFile.Close(); err != nil {
+		log.Fatalf("Failed to close ledger file: %v", err)
 	}
 	// Metadata about block production
 	var chainHeight = youngestBlockHeader.Height
@@ -154,12 +154,12 @@ func main() {
 					log.Fatalf("failed to open ledger file: %s\n", err)
 				}
 				err = blockchain.AddBlock(newBlock, blockchainFile, metadataDatabaseConnection)
-				if err := blockchainFile.Close(); err != nil {
-					log.Fatalf("Failed to close blockchain file: %v", err)
-				}
 				if err != nil {
 					log.Fatalf("Failed to add block %v", err)
 				} else {
+					if err := blockchainFile.Close(); err != nil {
+						log.Fatalf("Failed to close blockchain file: %v", err)
+					}
 					chainHeight++
 					// Reset pending pool map to empty
 					for k := range pendingMap.Sender {
--- a/cmd/handlers/main.go
+++ b/cmd/handlers/main.go
@@ -158,6 +158,9 @@ func main() {
 					// Reset pool
 					pendingContractPool = nil
 
+					// Reset youngest block header
+					youngestBlockHeader = newBlock.GetHeader()
+
 					// Reset production interval
 					go triggerInterval(intervalChannel, productionInterval)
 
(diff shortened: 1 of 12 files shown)


cmd/handlers/main.go
```diff
@@ -136,7 +149,15 @@ func main() {
 			} else {
 
 				// Add block to blockchain
-				if err := blockchain.AddBlock(newBlock, constants.BlockchainFile, constants.MetadataTable); err != nil {
+				blockchainFile, err := os.OpenFile(constants.BlockchainFile, os.O_APPEND|os.O_WRONLY, 0644)
```
kstola2: Why is this being opened a second time?

HarryL5004: The file is opened with different flags here for the purpose of appending new data, whereas when the file is first opened it was simply for read only.

HarryL5004: The `(*FILE) seek` function within the `getBlockBy...` funcs does not work as intended when a file is opened with the flag `O_APPEND`. That's why I chose to open the files for `addBlock` and `getBlockBy...` separately.

kstola2: My concern is that the second time we open the file, we are opening it within a loop, so it will happen each time. Is it possible to close the file pointer the first time once we're done, then open a new one for appending, and just reuse that for the loop?

cmd/handlers/main.go
```diff
@@ -136,7 +149,15 @@ func main() {
...
+				if err != nil {
+					log.Fatalf("failed to open ledger file: %s\n", err)
+				}
+				err = blockchain.AddBlock(newBlock, blockchainFile, metadataDatabaseConnection)
+				if errClosing := blockchainFile.Close(); errClosing != nil {
+					log.Fatalf("Failed to close blockchain file: %v", err)
```
kstola2: Why is `errClosing` being declared if we print out `err` in the `log.Fatalf`?

HarryL5004: Oh, `log.Fatalf` should print `errClosing` rather than `err` here.

cmd/handlers/main.go
HarryL5004: It is certainly possible to reuse the same file pointer for appending. I was concerned about opening a file indefinitely.

kstola2: I don't feel there's anything wrong with opening the file indefinitely. Although I suppose it might become an issue if we want to read from it. We can keep it as is unless it becomes an issue later.

cmd/handlers/main.go
kstola2: Instead of `errClosing` you can just call it `err`. It will be a different `err` than the one on line 156, because its within its own scope. That one should also be handled though.

- [ ] handle `AddBlock` `err`
- [ ] use `err` instead of `errClosing` in `if` block

kstola2: Shouldn't we handle the `AddBlock` error first? It seems if it is not nil, we log fatally anyway. If it is nil, then in the else block starting on line 162 we can then close it, and handle that error.

- [ ] Handle `AddBlock` error first; if it works correctly, handle the `Close` error

cmd/handlers/main.go
```diff
@@ -77,11 +84,17 @@ func main() {
...
+	ledgerFile, err := os.OpenFile(constants.BlockchainFile, os.O_RDONLY, 0644)
+	if err != nil {
+		log.Fatalf("Failed to open ledger file")
+	}
+	youngestBlockHeader, err := blockchain.GetYoungestBlockHeader(ledgerFile, metadataDatabaseConnection)
 	if err != nil {
 		log.Fatalf("Failed to get youngestBlockHeader")
 	}
-
+	if errClosing := ledgerFile.Close(); errClosing != nil {
```
kstola2: - [ ] Change `errClosing` to `err` and add it to the log statement.

Issue #228, opened 2019-06-19
Add error returns to keys functions.


kstola2:
Currently the encoding & decoding functions in the `keys` package are simply ignoring errors within the function body instead of returning them if they are faulty. If a conversion fails we should know about it and handle it gracefully instead of panicking.

This may cause some tests in the project overall to fail because of their current usage.

Example of area where an error needs to be handled: https://github.com/SIGBlockchain/project_aurum/blob/dev/internal/publickey/publickey.go#L23

- [x] all `privatekey` and `publickey` package functions handle errors if possible
- [x] all `privatekey` and `publickey` package tests pass
- [x] all usages of modified `privatekey` and `publickey` functions in repo refactored to handle errors
- [x] Use commit history to identify and validate changes

bashbashbash:
via #306, this issue was resolved

ramirez915:
Changes on the WIP regarding `publickey.Encode()` and `publickey.Decode()` for error handling look good to me and the changes throughout the project to account for these changes make sense to me.

ramirez915:
In `privatekey.GenerateNRandomKeys()` on line 80, the line currently is `x509Encoded, _ := x509.MarshalECPrivateKey(p)` but the function `x509.MarshalECPrivateKey()` returns an error that is not handled. I think this should be changed to handle the error it returns but I'm not entirely sure. Aside from this, all tests pass.

kstola2:
@ramirez915 Do we use that function anywhere? The `GenerateNRandomKeys`

ramirez915:
I pulled from dev to check and did not find the `GenerateNRandomKeys` used anywhere in the project besides in privatekey_test.go file

ramirez915:
I created a card regarding `GenerateNRandomKeys` in the backlog. Since `GenerateNRandomKeys` is not currently used anywhere in the project besides in privatekey_test.go the handling for the `x509.MarshalECPrivateKey()` can be ignored for now.

PR #295, opened 2019-07-29
Add error returns to keys functions.#228

bashbashbash:
Did not include errors for the Equals function because boolean type functions inherently have a result that suggests success or failure. Am willing to discuss.


bashbashbash:
This branch was accidentally merged. The issue will be resolved in a future PR

PR #296, opened 2019-07-29
Change accounts package I/O function file parameter

HarryL5004:
Issue #250

### DIFF
--- a/internal/producer/src/validation/validation.go
+++ b/internal/producer/src/validation/validation.go
@@ -3,10 +3,12 @@ package validation
 import (
 	"bytes"
 	"crypto/ecdsa"
+	"database/sql"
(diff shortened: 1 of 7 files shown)


internal/producer/src/validation/validation.go
```diff
@@ -48,8 +50,15 @@ func ValidateContract(c *contracts.Contract) error {
 	}
 
 	// retrieve sender's balance from account balance table
+	dbConnection, err := sql.Open("sqlite3", constants.AccountsTable)
```
kstola2: Can we pass this in as a parameter instead?

HarryL5004: sure

PR #299, opened 2019-07-29
Fracture producer pkg

ramirez915:
Moved functions around to specified packages, packages to internal and the Dockerfile into root


internal/accountstable/accountstable.go
```diff
@@ -1,12 +1,18 @@
...
+	"bytes"
 	"database/sql"
 	"encoding/hex"
 	"errors"
 	"fmt"
 
-	"github.com/SIGBlockchain/project_aurum/internal/producer/src/accountinfo"
+	"github.com/SIGBlockchain/project_aurum/internal/accountinfo"
+	"github.com/SIGBlockchain/project_aurum/internal/block"
+	// "github.com/SIGBlockchain/project_aurum/internal/constants"
```
kstola2: - [ ] Remove comment

ramirez915: my bad, got it now

ramirez915: I found some sql statements that are new that need to use the constants. Currently changing that

Issue #272, opened 2019-07-03
Fracture client package


kstola2:
Fracture the client package into `internal/wallet`. This new package should contain:

- `SetupWallet`
- `GetPrivateKey`
- `GetBalance`
- `GetStateNonce`
- `GetWalletAddress`
- `RequestWalletInfo`
- `UpdateWallet`

Any functions/tests that don't make the cut should be removed along with the original `client` package.

The package should reside in `internal/`

- [x] New package `internal/wallet` created
- [x] Contains only functions mentioned above
- [x] Tests that support above functions are present in `wallet_test.go`
- [x] `client` root directory and all subdirectories deleted

HarryL5004:
Completed

PR #304, opened 2019-08-02
Fracture client package

HarryL5004:
Issue #272


internal/wallet/wallet.go
```diff
@@ -28,89 +19,7 @@ import (
 
 var SecretBytes = hashing.New([]byte("aurum"))[8:16]
```
kstola2: Do we use this anymore anywhere?

HarryL5004: It is used in `RequestWalletInfo`.

kstola2: That's my mistake. I don't think we need that function anymore since we're using `http`

PR #310, opened 2019-08-08
cmd main.go

ramirez915:
Handler's version of main.go is in cmd

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -166,8 +166,13 @@ func main() {
 					}
 					// Update accounts table with all contracts in pool
 					for _, contract := range pendingContractPool {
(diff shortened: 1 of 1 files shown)


kstola2:
those two keys that are being hashed in the log statement around 139 just need to be initialized prior to log statement and then passed in as variables @ramirez915

ramirez915:
@kstola2 yea I noticed after working with it. I started from the bottom which I now know not to do because I ended up deleting that code. At least I had the right idea. I think it's all fixed now. I ran main.go and was able to produce blocks according to the terminal. There were 0 contracts but I didn't seem to get any errors

Issue #322, opened 2019-09-03
Setting up aurum docker package


kstola2:
In an effort to make more comprehensive testing on Travis, new types of tests will be created called integration tests. These tests will basically run docker containers and spin up a producer and maybe a few clients, interact in some meaningful way, and then make assertions based on some activity. An example would be spinning up a container with a producer and checking to make sure it does an airdrop correctly. These types of tests deal with _systemic behavior_ after a potential code change has been made, as opposed to a unit test which simply tests one particular feature to make sure it's working.

For the first issue dealing with this, create a aurum docker package in internal that has a library file and a test file just like every other package. Naming is up to you.

As a first test simply copy the code from the container run example [here](https://docs.docker.com/develop/sdk/examples/#run-a-container) and put it in a test function. If it runs without problems, then it is complete.

HarryL5004:
Completed

PR #326, opened 2019-09-08
Setting up aurum docker package

HarryL5004:
Issue #322


kstola2:
Can you change the destination branch to dev? p2 is the "phase 2 master".

Issue #321, opened 2019-09-03
Settings commands


kstola2:
Currently `config.json` has the following structure:
```
{
    "Version": 1,
    "InitialAurumSupply": 50000000,
    "Port": "12000",
    "BlockProductionInterval": "20s",
    "Localhost": true
}
```

In an effort to make changes easier, create a separate `main.go` which will be a produce a binary that can change a particular configuration setting depending on what flag is specified.

So theoretically you would be able to run `someBinary -port 35000` and that would go into the `config.json` file and change the port.

Resource: https://golang.org/pkg/flag/

HarryL5004:
Completed

PR #327, opened 2019-09-09
Settings Commands

HarryL5004:
Issue #321

### DIFF
--- a/config/main.go
+++ b/config/main.go
@@ -1,0 +1,64 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"io/ioutil"
+	"log"
+	"os"
+	"strings"
+
+	"github.com/SIGBlockchain/project_aurum/internal/config"
+)
+
+func main() {
+	//open configuration file
+	configFile, err := os.Open("../cmd/config.json")
+	if err != nil {
+		log.Fatal("Failed to load configuration file : " + err.Error())
+	}
+	defer configFile.Close()
+
+	cfgData, err := ioutil.ReadAll(configFile)
+	if err != nil {
+		log.Fatal("Failed to read configuration file : " + err.Error())
+	}
+
+	cfg := config.Config{}
+	if err := json.Unmarshal(cfgData, &cfg); err != nil {
+		log.Fatal("Failed to unmarshall configuration data : " + err.Error())
+	}
+
+	//specify flags
+	versionU64 := flag.Uint64("version", uint64(cfg.Version), "enter version number")
+	flag.Uint64Var(&cfg.InitialAurumSupply, "initialaurumsupply", cfg.InitialAurumSupply, "enter a number for initial aurum supply")
+	flag.StringVar(&cfg.Port, "port", cfg.Port, "enter port number")
+	flag.StringVar(&cfg.BlockProductionInterval, "blockproductioninterval", cfg.BlockProductionInterval, "enter a time for block production interval")
+	flag.BoolVar(&cfg.Localhost, "localhost", cfg.Localhost, "enter either true/false")
+
+	//read flags
+	flag.Parse()
+
+	cfg.Version = uint16(*versionU64)
+	//check block production interval suffix
+	hasSuf := false
+	for _, s := range [7]string{"ns", "us", "µs", "ms", "s", "m", "h"} {
+		if strings.HasSuffix(cfg.BlockProductionInterval, s) {
+			hasSuf = true
+			break
+		}
+	}
+	if !hasSuf {
+		cfg.BlockProductionInterval += "s"
+	}
+
+	//write into configuration file
+	marshalledJSON, err := json.Marshal(cfg)
+	if err != nil {
+		log.Fatalf("Failed to marshal new config: %v", err)
+	}
+
+	if err := ioutil.WriteFile("../cmd/config.json", marshalledJSON, 0644); err != nil {
+		log.Fatalf("failed to write to file: %v", err)
+	}
+}
--- a/settings/main.go
+++ b/settings/main.go
@@ -9,32 +9,29 @@ import (
 	"strings"
 
 	"github.com/SIGBlockchain/project_aurum/internal/config"
+	"github.com/SIGBlockchain/project_aurum/internal/jsonify"
 )
 
 func main() {
 	//open configuration file
 	configFile, err := os.Open("../cmd/config.json")
 	if err != nil {
-		log.Fatal("Failed to load configuration file : " + err.Error())
+		log.Fatal("Failed to open configuration file : " + err.Error())
 	}
 	defer configFile.Close()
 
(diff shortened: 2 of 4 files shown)


config/main.go
```diff
@@ -0,0 +1,64 @@
...
+	if err := json.Unmarshal(cfgData, &cfg); err != nil {
+		log.Fatal("Failed to unmarshall configuration data : " + err.Error())
+	}
+
+	//specify flags
+	versionU64 := flag.Uint64("version", uint64(cfg.Version), "enter version number")
+	flag.Uint64Var(&cfg.InitialAurumSupply, "initialaurumsupply", cfg.InitialAurumSupply, "enter a number for initial aurum supply")
+	flag.StringVar(&cfg.Port, "port", cfg.Port, "enter port number")
+	flag.StringVar(&cfg.BlockProductionInterval, "blockproductioninterval", cfg.BlockProductionInterval, "enter a time for block production interval")
+	flag.BoolVar(&cfg.Localhost, "localhost", cfg.Localhost, "enter either true/false")
```
kstola2: The `localhost` flag doesn't appear to be working. I tried `./config -localhost false` and `./config -localhost 0` and it didn't change.

config/main.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	defer configFile.Close()
+
+	cfgData, err := ioutil.ReadAll(configFile)
+	if err != nil {
+		log.Fatal("Failed to read configuration file : " + err.Error())
+	}
+
+	cfg := config.Config{}
+	if err := json.Unmarshal(cfgData, &cfg); err != nil {
```
kstola2: I believe we have a function that does the configuration loading. I believe it's in `internal/config/config.go`.

config/main.go
```diff
@@ -0,0 +1,64 @@
...
+		log.Fatal("Failed to unmarshall configuration data : " + err.Error())
+	}
+
+	//specify flags
+	versionU64 := flag.Uint64("version", uint64(cfg.Version), "enter version number")
+	flag.Uint64Var(&cfg.InitialAurumSupply, "initialaurumsupply", cfg.InitialAurumSupply, "enter a number for initial aurum supply")
```
kstola2: I think it would be easier on the user to make some of the long flags be single words.

config/main.go
```diff
@@ -0,0 +1,64 @@
...
+		log.Fatal("Failed to read configuration file : " + err.Error())
+	}
+
+	cfg := config.Config{}
+	if err := json.Unmarshal(cfgData, &cfg); err != nil {
+		log.Fatal("Failed to unmarshall configuration data : " + err.Error())
+	}
+
+	//specify flags
+	versionU64 := flag.Uint64("version", uint64(cfg.Version), "enter version number")
```
kstola2: Not a big deal, but I think using a `flag.Uint` would be more appropriate than a 64-bit number, simply because we're eventually casting down to a 16-bit, so 32-bit might look better? idk

config/main.go
```diff
@@ -0,0 +1,64 @@
...
+	//check block production interval suffix
+	hasSuf := false
+	for _, s := range [7]string{"ns", "us", "µs", "ms", "s", "m", "h"} {
+		if strings.HasSuffix(cfg.BlockProductionInterval, s) {
+			hasSuf = true
+			break
+		}
+	}
+	if !hasSuf {
+		cfg.BlockProductionInterval += "s"
```
kstola2: I like what you did here, makes it much more convenient.

kstola2:
Should we rename it from `config` since there's already an `internal/config` directory?

config/main.go
HarryL5004: Oh that's because the syntax `-flag x` is for non-boolean flags only. The flag works if you use the following syntax: `-flag=x`. I can change it to accept the other syntax too if you like.

config/main.go
HarryL5004: The problem with that function is that it doesn't have the correct path to the `json` file, since this config file is in another directory.

config/main.go
kstola2: I think keeping it simple is best, but make sure to include the correct syntax in the help string.

config/main.go
kstola2: How about the `jsonify` package? It doesn't have a test to support it yet but maybe that would work?

config/main.go
HarryL5004: I thought `flag.Uint` returns a pointer to a 8-bit uint? That's why I used the `Uint64` version...

kstola2: Integers are typically 32 bit. 

https://golang.org/pkg/builtin/#uint

settings/main.go
```diff
@@ -0,0 +1,61 @@
...
+
+	cfg := config.Config{}
+	err = jsonify.LoadJSON(configFile, &cfg)
+	if err != nil {
+		log.Fatal("Failed to unmarshall configuration data : " + err.Error())
+	}
+
+	//specify flags
+	versionU64 := flag.Uint("version", uint(cfg.Version), "enter version number")
+	flag.Uint64Var(&cfg.InitialAurumSupply, "initialaurum", cfg.InitialAurumSupply, "enter a number for initial aurum supply")
```
kstola2: - [ ] Let's just rename this to `supply`

settings/main.go
```diff
@@ -0,0 +1,61 @@
...
+	flag.StringVar(&cfg.Port, "port", cfg.Port, "enter port number")
+	flag.StringVar(&cfg.BlockProductionInterval, "blockproditv", cfg.BlockProductionInterval, "enter a time for block production interval\n(assuming seconds if units are not provided)")
```
kstola2: - [ ] Rename to `interval`

settings/main.go
```diff
@@ -0,0 +1,61 @@
...
+	flag.BoolVar(&cfg.Localhost, "localhost", cfg.Localhost, "syntax: -localhost=/boolean here/")
+
+	//read flags
+	flag.Parse()
+
+	cfg.Version = uint16(*versionU64)
+	//check block production interval suffix
+	hasSuf := false
+	for _, s := range [7]string{"ns", "us", "µs", "ms", "s", "m", "h"} {
+		if strings.HasSuffix(cfg.BlockProductionInterval, s) {
```
kstola2: Can you include in the instructions that it defaults to seconds? Or a simpler solution would be to just say "invalid suffix" if the suffix doesn't exist in the list here.

kstola2:
I believe Go convention wants all main files in `cmd/`, so can we put this `settings/` package inside `cmd/`?

settings/main.go
HarryL5004: I've included a hint in the instructions that says that it defaults to seconds.
`33: flag.StringVar(&cfg.BlockProductionInterval, "blockproditv", cfg.BlockProductionInterval, "enter a time for block production interval\n(assuming seconds if units are not provided)")`

PR #329, opened 2019-09-10
completed issue 324

bashbashbash:
Even though the tests have been consolidated, I would like to implement the tests in a slightly better way using a struct, something like this from another test:

My question is, would this be a cleaner way? I think so but don't want to waste time

tests := []struct {
		input [][]byte
		want  bool
	}{
		{
			[]byte{something},
			true,
		},

                  ...
}

### DIFF
--- a/internal/hashing/hashing_test.go
+++ b/internal/hashing/hashing_test.go
@@ -20,18 +20,16 @@ func TestNew(t *testing.T) {
 	}
 }
 
-func TestGetMerkleRootHashEmptyInput(t *testing.T) {
+func TestGetMerkleRootHashInput(t *testing.T) {
 	input := [][]byte{}
-	result := GetMerkleRootHash(input)
+	expected := GetMerkleRootHash(input)
 
-	if len(input) != len(result) {
+	if len(input) != len(expected) {
 		t.Errorf("Error! GetMerkelRootHash does not return an empty slice on input of empty slice")
 	}
-}
 
-func TestGetMerkleRootHashSinlgeInput(t *testing.T) {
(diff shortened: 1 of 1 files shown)


internal/hashing/hashing_test.go
```diff
@@ -20,65 +20,54 @@ func TestNew(t *testing.T) {
...
-	input := [][]byte{[]byte("transaction1"), []byte("transaction2")}
-	concat := append(New(New(input[0])), New(New(input[1]))...)
-	expected := New(New(concat))
-	actual := GetMerkleRootHash(input)
+func TestGetMerkleRootHashInput(t *testing.T) {
+	tests := []struct {
+		name string
+		input [][]byte
+		expected []byte
+		want bool
```
kstola2: what is the purpose of `want` here?

internal/hashing/hashing_test.go
```diff
@@ -20,65 +20,54 @@ func TestNew(t *testing.T) {
...
-	expected := New(New(append(concat1, concat2...)))
-	actual := GetMerkleRootHash(input)
-
-	if !bytes.Equal(expected, actual) {
-		t.Errorf("Error! GetMerkelRootHash does not produce correct result on three byte slices")
-		t.Errorf("Expected != Actual")
-		t.Errorf("%v != %v", expected, actual)
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if result := bytes.Equal(GetMerkleRootHash(tt.input),tt.expected); result != tt.want {
```
kstola2: can we just make the assertion fail if GetMerkleRootHash bytes do not equal expected?

Issue #325, opened 2019-09-03
Go routine in TestContractRequestHandler


kstola2:
Why is there a need for a `go` routine here? Can we make this test pass without it? 

The line in question: https://github.com/SIGBlockchain/project_aurum/blob/master/internal/handlers/handlers_test.go#L295

HarryL5004:
I think the `go` routine is there is because there's a contract channel in the test and in `HandleContractRequest` that waits for each other before channeling a contract.

kstola2:
@HarryL5004 I just noticed I completely linked the wrong thing

kstola2:
> I think the `go` routine is there is because there's a contract channel in the test and in `HandleContractRequest` that waits for each other before channeling a contract.

I see what you mean now. Is there a way to make this test pass without the sleep call?

HarryL5004:
It seems the `handler` needs at least a nanosecond to process the request before it can return the correct HTTP status code and response. Perhaps it's due to all of the processes leading up to contract validation in `pendingMap.Add()`?

A nanosecond sleep call seems to work just fine.

kstola2:
The problem with that solution is it might be a nanosecond on your machine but on travis or a VM or someone else's machine it might not be enough. It's not a huge deal, but if this is to be a learning point, we need to think about code that runs as fast as the machine allows it. I think instead of sleeping, it might be best to leverage a `WaitGroup` and add the `goroutine` to it, then wait for it to finish before proceeding:

https://golang.org/pkg/sync/#WaitGroup

What do you think?

HarryL5004:
> The problem with that solution is it might be a nanosecond on your machine but on travis or a VM or someone else's machine it might not be enough. It's not a huge deal, but if this is to be a learning point, we need to think about code that runs as fast as the machine allows it. I think instead of sleeping, it might be best to leverage a `WaitGroup` and add the `goroutine` to it, then wait for it to finish before proceeding:
> 
> https://golang.org/pkg/sync/#WaitGroup
> 
> What do you think?

Right, for some reason I wasn't thinking about other environments/machines.

I was thinking about adding `WaitGroup` as well. I will try to do that.

HarryL5004:
Completed

PR #341, opened 2019-09-25
Go routine in TestContractRequestHandler

HarryL5004:
Issue #325

### DIFF
--- a/internal/handlers/handlers_test.go
+++ b/internal/handlers/handlers_test.go
@@ -13,7 +13,6 @@ import (
 	"os"
 	"sync"
 	"testing"
-	"time"
 
 	"github.com/SIGBlockchain/project_aurum/internal/accountstable"
 	"github.com/SIGBlockchain/project_aurum/internal/constants"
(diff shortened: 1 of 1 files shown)


kstola2:
What happened to using WaitGroup?

HarryL5004:
If waitgroup is used in `cmd main.go`, the waitgroup counter can go negative and panic.

kstola2:
Why would we news one in main.go? The issue just concerns the test function.

HarryL5004:
That's because a wait group instance must be passed into the contract handler to signal the waitgroup that this goroutine is done, and main.go initiate a contract handler.

kstola2:
Hm. Can you do something like 

```
var waitgroup sync.WaitGroup
waitgroup.Add(1)
go func(){
    handler.ServeHTTP(rr, tt.req)
    waitgroup.Done()
}()
waitgroup.Wait()
```

kstola2:
I don't think it's a great idea to edit so many files for this issue, I'd almost rather just have the 1 second sleep at that point. But there's gotta be a way to use a WaitGroup to achieve this.

kstola2:
Interesting, do you know why increasing the cap makes it work? I'm curious

HarryL5004:
Previously the channel blocks all processes until there's a receiver retrieving the contract in the channel. By increasing the capacity, the channel will only start to block processes when it is full. If the channel is never full, it will block any processes.

PR #342, opened 2019-09-26
Docker test

kstola2:
Will now spin up a VM with project aurum on it, allowing for interaction in tests


Vineet77:
Is it ok to have a binary?

kstola2:
Good point. So I included in travis a thing that forces you to rebuild the binary before it can be approved. That way the binary on dev is always up to date

kstola2:
@Vineet77 I removed the binary check for now from travis, and I'll work on making some sort of binary-release scripting issue.

Issue #344, opened 2019-09-30
Block Validation


HarryL5004:
Package context: `internal/block`
Branch: `Block_Validation`

For this issue, add code to the `validate` function stub in `internal/block`. 
`Validate` should return **false** if:

- Block version does not match the version given in the parameter
- Block height does not match the height given in the parameter
- Block previous hash does not match the previous hash given in the parameter
- Block timestamp is before the timestamp given in the parameter
- Block timestamp is in the future

A test for this function is provided in the test file of package `block`.

Make sure to leverage the existing functions in `internal/hashing` and/or `internal/block`.

kstola2:
@HarryL5004 wouldn't it make more sense to put this in the `validation/` package?

PR #349, opened 2019-10-01
Block validation function

arosso3:
Issue: #344


HarryL5004:
Putting the function into `internal/validate` does make more sense. I guess I forgot about the existence of the validation package. I thought validateContract is in the package contract.

internal/block/block_test.go
```diff
@@ -395,3 +395,113 @@ func TestHashBlockHeader(t *testing.T) {
...
+			false,
+		},
+		{
+			"Invalid timestamp",
+			Block{
+				Version:        baseBlk.Version,
+				Height:         baseBlk.Height,
+				PreviousHash:   baseBlk.PreviousHash,
+				MerkleRootHash: baseBlk.MerkleRootHash,
+				Timestamp:      baseBlk.Timestamp + 1,
```
kstola2: Shouldn't this timestamp be before the parameter?

internal/validation/validation.go
```diff
@@ -133,3 +134,26 @@ func ValidatePending(c *contracts.Contract, pBalance *uint64, pNonce *uint64) er
...
+	// Check Height
+	if b.Height != height {
+		return false
+	}
+	// Check Previous Hash
+	if !(bytes.Equal(b.PreviousHash, previousHash)) {
+		return false
+	}
+	// Check timestamp
+	if b.Timestamp != timeStamp {
```
kstola-bloxroute: This doesn't allow for timestamps that have been mined between "now" and the parameter timestamp.

internal/validation/validation_test.go
```diff
@@ -317,3 +319,113 @@ func TestValidatePending(t *testing.T) {
...
+			false,
+		},
+		{
+			"Invalid timestamp (future)",
+			block.Block{
+				Version:        baseBlk.Version,
+				Height:         baseBlk.Height,
+				PreviousHash:   baseBlk.PreviousHash,
+				MerkleRootHash: baseBlk.MerkleRootHash,
+				Timestamp:      baseBlk.Timestamp + 1,
```
kstola-bloxroute: This can be a valid timestamp, it just can't be in the future from "now", where "now" is the point at which the validation is called.

Issue #346, opened 2019-10-01
Endpoint for Handling Queries by Height and by Hash


bashbashbash:
**Workspace:**
- `internal/endpoints/endpoint.go`

Create a  constant in `endpoints.go` that will be used to accept incoming blocks by height. The endpoint should be `/block/height`

Create a constant in 'endpoint.go' that will be used to accept incoming blocks by hash. The endpoint should be `/block/hash`

kstola2:
https://github.com/SIGBlockchain/project_aurum/blob/master/internal/endpoints/endpoints.go

kstola2:
@bashbashbash What do you mean by "keys" here?

bashbashbash:
> @bashbashbash What do you mean by "keys" here?

Good point - meant values and updated the issue statement

harshalpatel165:
completed

PR #355, opened 2019-10-04
Additionalconstants

harshalpatel165:
(#346 ) added constants for incoming blocks by height and hash with endpoints /block/height and /block/hash


harshalpatel165:
ok im not sure how to fix this

Issue #360, opened 2019-10-08
Complete GetBlockByHashRequest function


bashbashbash:
Package context: `internal/request`
Branch: none

Create test and function that returns a request for a block given a block hash

**Resources:**
- [HTTP Requests](https://github.com/SIGBlockchain/project_aurum/wiki/HTTP-Requests)

bashbashbash:
#361 Complete

PR #361, opened 2019-10-08
Getblockbyhashrequest

bashbashbash:
Preliminary submission for #360

### DIFF
--- a/internal/requests/requests.go
+++ b/internal/requests/requests.go
@@ -62,3 +62,7 @@ func GetBlockByHeightRequest(blockHeight uint64) (*http.Request, error) {
 	req.URL.RawQuery = values.Encode()
 	return req, nil
 }
+
+func GetBlockByHashRequest(blockHash string) (*http.Request, error) {
+	return nil, errors.New("Failed to make a new request")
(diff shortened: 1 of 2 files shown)


internal/requests/requests.go
```diff
@@ -62,3 +62,14 @@ func GetBlockByHeightRequest(blockHeight uint64) (*http.Request, error) {
...
 	return req, nil
 }
+
+func GetBlockByHashRequest(blockHash string) (*http.Request, error) {
+	req, err := http.NewRequest(http.MethodGet, endpoints.BlockQueryByHash, nil)
+	if err != nil {
+		return nil, errors.New("Failed to make new request:\n" + err.Error())
+	}
+	values := req.URL.Query()
+	values.Add("#", blockHash)
```
kstola2: I'm not sure if non-alphabetical keys are allowed in http urls, can you check?

bashbashbash: Updating to p for pound (I think I would have to use the ascii code for '#', which I don't want to do)

bashbashbash:
#361

Issue #350, opened 2019-10-01
Merkle-Root Hash Comparison


HarryL5004:
Package context: `internal/hashing`
Branch: `MerkleRootcmp`

For this issue, add code to the function stub `MerkleRootHashOf` in `internal/hashing`.
The function should return **false** if the given merkle-root hash is not the merkle root hash of the array of SHA-256 hashes given in the parameter.

A test for this function is provided in the test file of package `hashing`.

HarryL5004:
Completed

PR #364, opened 2019-10-08
Merkle-Root Hash Comparison

HarryL5004:
Issue #350


kstola2:
Sorry I forgot to ask, can there also be a test case for this in Block Validation?

HarryL5004:
Sorry I forgot to add a test case for it in validationblock.

Issue #318, opened 2019-08-30
Testing: jsonify


bashbashbash:
package context: internal/jsonify

Please create tests for the function LoadJSON. Please consider cases where interface is not empty, interface is invalid, interface is nil, file is nil, file is invalid, file is not open, and how to handle various errors returned. Also consider valid cases.

Please following file naming conventions for testing and please include comments to specify what each test is performing

**Resources:**
- [Go Testing Package](https://golang.org/pkg/testing/)
- [Go Logging Package](https://golang.org/pkg/log/) (for detailed debug printing)
- [Go reflect Package](https://golang.org/pkg/reflect/) (for comparison)

harshalpatel165:
does a file being nil mean that that the file does not exist or that there is nothing within the file

harshalpatel165:
tbh, I'm not sure what I'm supposed to be doing with this. Do I have to create an interface in the test and run the LoadJSON function with it. If so, what goes in the interface.

harshalpatel165:
completed

PR #367, opened 2019-10-09
Testing json

harshalpatel165:
(#318 ) this is what I have gotten so far. I dont know if I'm on the right track. I created a temp file which will be closed and deleted after all the tests. I also created an empty interface. I used these two to test if the file is invalid and the interface is nil.


kstola2:
Use the other PR.

Issue #377, opened 2019-10-13
New function for public key package


kstola2:
- [ ] Create a struct called `AurumPublicKey` with `Key`, `Bytes`, `Hex`, `Hash`,  fields
- [ ] Create a function called `New` that takes in a `*ecdsa.PrivateKey` as a parameter and generates an `AurumPublicKey`
- [ ] `Key` fields should be set to the private key's public key like here: https://github.com/SIGBlockchain/project_aurum/blob/master/internal/publickey/publickey_test.go#L83
- [ ] `Bytes` field should be the return of calling `publicKey.Encode` with the `Key` as the parameter
- [ ] `Hex` should be set to the return of `hex.EncodeToString` with `Bytes` as the parameter
- [ ] `Hash` should be set to the return of `hashing.New` with `Bytes` field as a parameter
- [ ] Test should just check that each field is set according to above requirements

Vineet77:
Resolved by https://github.com/SIGBlockchain/project_aurum/pull/390

Issue #332, opened 2019-09-22
Add posting to blockpub goroutine


kstola2:
_Description_: BlockPub is a page owned by the ACM used by SIG Blockchain as a "meeting place" for active network users of Aurum. The idea here is to create a "heartbeat" for the producer, periodically advertising its IP:Port combo publicly to say its alive and accepting requests. 

- [ ] Call `nettools.GetIP` in `main.go` (right after the call to `http.ListenAndServe`
- [ ] Call `requests.AddPeerToDiscoveryRequest` and send it off
- [ ] Check http://blockchain.acmapp.tech/peers (under Raw Data) to see if the information is populated in the array (**Please screenshot this to show that it's working in your PR**)
- [ ] Make a goroutine that repeats `requests.AddPeerToDiscoveryRequest` send every 5 minutes

Vineet77:
http://blockchain.acmapp.tech/peers is currently not working

Vineet77:
Resolved by https://github.com/SIGBlockchain/project_aurum/pull/391

PR #390, opened 2019-10-21
Created New function for AurumPublicKey

Vineet77:
Created a New function, corresponding to issue: https://github.com/SIGBlockchain/project_aurum/issues/377

Also added unit tests (that pass)


kstola2:
What's up with Issue #332 ?

PR #391, opened 2019-10-21
Posting to peer discovery list

Vineet77:
Main.go will now launch a gorotuine that will send requests to http://blockchain.acmapp.tech/peers to add the producer to the list of peers.

image

Proof ^

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -12,6 +12,9 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/SIGBlockchain/project_aurum/internal/nettools"
+	"github.com/SIGBlockchain/project_aurum/internal/requests"
+
 	"github.com/SIGBlockchain/project_aurum/internal/accountstable"
 	"github.com/SIGBlockchain/project_aurum/internal/block"
 	"github.com/SIGBlockchain/project_aurum/internal/blockchain"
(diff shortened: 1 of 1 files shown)


cmd/main.go
```diff
@@ -209,3 +218,21 @@ func triggerInterval(intervalChannel chan bool, productionInterval time.Duration
...
+	if err != nil {
+		log.Fatalf("Failed to make Add Peer request: %s", err.Error())
+	}
+	client.Do(r)
+	for range time.Tick(time.Second * 300) { //5 minutes
+		r, err := requests.AddPeerToDiscoveryRequest(ip, port)
+		if err != nil {
+			log.Fatalf("Failed to make Add Peer request: %s", err.Error())
+		}
+		client.Do(r)
```
kstola2: Can we handle the `Do` error and just log it if it fails? Not fatally but just log it in case the peer server goes down.

cmd/main.go
```diff
@@ -209,3 +218,21 @@ func triggerInterval(intervalChannel chan bool, productionInterval time.Duration
 	time.Sleep(productionInterval)
 	intervalChannel <- true
 }
+
+func addPeertoDiscoveryList(ip, port string) {
+	client := http.Client{}
+	//send initally and then wait every 5 minutes
+	r, err := requests.AddPeerToDiscoveryRequest(ip, port)
+	if err != nil {
+		log.Fatalf("Failed to make Add Peer request: %s", err.Error())
```
kstola2: Let's not fatally exit if it fails to make the request. It should still be able to function.

Vineet77: I'll change to to a normal log, and continually try to send.

cmd/main.go
Vineet77: Done

Issue #379, opened 2019-10-13
Settings main.go bug


kstola2:
Bug with settings, if I type `1hr` the config changes to `1hrs`. I think it's appending `s` to anything that isn't a valid time-interval. Should reject anything that isn't a valid time interval, and maybe output valid time intervals if the argument is invalid.

bashbashbash:
Approved and closed #392

PR #392, opened 2019-10-22
completed issue 379

bashbashbash:
Completed issue #379

### DIFF
--- a/cmd/settings/main.go
+++ b/cmd/settings/main.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"regexp"
 	"strings"
 
 	"github.com/SIGBlockchain/project_aurum/internal/config"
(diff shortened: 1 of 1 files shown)


cmd/settings/main.go
```diff
@@ -37,16 +38,30 @@ func main() {
...
+	validInterval, _ = regexp.MatchString("[a-s]", string([]rune(cfg.BlockProductionInterval)[len(cfg.BlockProductionInterval)-1]))
+	if !validInterval {
+		log.Fatalf("Interval input did not end with letter")
+	}
 	//check block production interval suffix
 	hasSuf := false
+
 	for _, s := range [7]string{"ns", "us", "µs", "ms", "s", "m", "h"} {
 		if strings.HasSuffix(cfg.BlockProductionInterval, s) {
 			hasSuf = true
```
kstola2: Instead of the `validInterval` above, can we instead leverage this function: https://golang.org/pkg/strings/#TrimSuffix

We can use it to validate that everything except for the suffix is a digit. The problem with `validInterval` is it would probably accept `1yns`, but I'm not sure cause there's no test for it.

bashbashbash: I ended up using TrimLeftFunc but I believe you shall enjoy and the validation should be more robust now. I made a card to clean up this file and add testing

kstola2: Sir I'm running `go run main.go` and the interval is not changing in the `config.json` file.

Let's make this validation a testable function in the `validation` package.

Issue #404, opened 2019-10-24
Implement HandleGetJSONBlockByHeight


bashbashbash:
Package context: `internal/handlers`
Branch: `getjsonblockbyheight`

Complete function HandleGetJSONBlockByHeight and pass test

kstola2:
Completed

PR #406, opened 2019-10-24
Fetchhandler

kstola2:
Issue #404


bashbashbash:
I have retested and everything is still fine

Issue #407, opened 2019-10-24
JSONBlock to Block


kstola2:
Create an Unmarshal function that converts a JSONBlock to a normal Block

bashbashbash:
Completed  and approved via #408

PR #408, opened 2019-10-25
finished function and test, but test not passing

bashbashbash:
Working on getting tests to pass - I believe the function has an error in converting string data to bytes

### DIFF
--- a/internal/block/block_test.go
+++ b/internal/block/block_test.go
@@ -458,7 +458,7 @@ func TestMarshal(t *testing.T) {
 	}
 }
 
-func TestUmarshal(t *testing.T) {
+func TestUnmarshal(t *testing.T) {
 	testBlock := Block{
 		Version:        3,
 		Height:         300,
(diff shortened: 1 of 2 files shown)


internal/block/block_test.go
```diff
@@ -457,3 +457,63 @@ func TestMarshal(t *testing.T) {
 		})
 	}
 }
+
+func TestUmarshal(t *testing.T) {
```
kstola-bloxroute: sp

internal/block/block_test.go
```diff
@@ -457,3 +457,63 @@ func TestMarshal(t *testing.T) {
...
+	testJSONBlock := JSONBlock{
```
kstola2: Can you call `block.Marshall` on a regular block and use that output as the `JSONBlock`? I just want to make sure it's consistent with that function.

internal/block/block_test.go
```diff
@@ -457,3 +457,69 @@ func TestMarshal(t *testing.T) {
 		})
 	}
 }
+
+func TestUmarshal(t *testing.T) {
```
kstola2: Just need this typo fixed

bashbashbash:
Completed #407

Issue #423, opened 2019-10-27
Generate Random Contract Function


kstola2:
I think we could accelerate development if we created a function that generates a random Contract. So in other words it returns a contract whose fields are all formed with stochasticity.

PR #426, opened 2019-10-30
added generaterandomcontract and containspublickey functions

harshalpatel165:
(#423 )

### DIFF
--- a/internal/contracts/contracts.go
+++ b/internal/contracts/contracts.go
@@ -9,7 +9,8 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
-
+	"math/rand"
+ 
 	"github.com/SIGBlockchain/project_aurum/internal/hashing"
 	"github.com/SIGBlockchain/project_aurum/internal/publickey"
 )
@@ -55,7 +56,7 @@ func New(version uint16, sender *ecdsa.PrivateKey, recipient []byte, value uint6
 
 	if version == 0 {
 		return nil, errors.New("Invalid version; must be >= 1")
-	}
+	}               
 
 	c := Contract{
 		Version:         version,
@@ -274,4 +275,53 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
 		mc.StateNonce,
 	}
 	return c, nil
+	fmt.print(c)
+}
+
+func (c *Contract) ContainsPublicKey(pk publickey.AurumPublicKey) (bool, error) {
+	if c == nil {
+		return false, errors.New("not nill")
+	}
+	
+	// encode contract's sender pub key to bytes
+	bKey, err := publickey.Encode(c.SenderPubKey)
+	if err != nil{
+		return bytes.Equal(bKey, pk.Bytes), errors.New("failed to encode sender public key")
+	}
+	// return the comparison of the above encoding with Bytes field of pk
+    return  bytes.Equal(bKey, pk.Bytes), nil
+	
+}
+
+func(c *Contract) GenerateRandomContract() (Contract){
+	b := rand.Read(make([]byte, 32))
+	
+	genSenderPubKey := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	
+	genrecipPubKeyHash := rand.Read(make([]byte, 32))
+	
+	genVersion := binary.LittleEndian.Uint16(b[:])
+	
+	genSigLen := binary.LittleEndian.Uint8(b[:]);
+	
+	
+	genSignature := rand.Read(make([]byte, genSigLen));
+	
+	genValue := binary.LittleEndian.Uint64(b[:])
+	
+	genStateNonce := binary.LittleEndian.Uint64(b[:])
+	
+
+	c := &Contract{
+		Version:  genVersion,
+		SenderPubKey: genSenderPubKey,
+		SigLen:  genSigLen,
+		Signature:  genSignature,
+		RecipPubKeyHash:  genrecipPubKeyHash,
+		Value:  genValue,
+		StateNonce:  genStateNonce,
+		
+	}
+	return c, nil
+
 }


internal/contracts/contracts.go
```diff
@@ -274,4 +275,53 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
 		mc.StateNonce,
 	}
 	return c, nil
+	fmt.print(c)
+}
+
+func (c *Contract) ContainsPublicKey(pk publickey.AurumPublicKey) (bool, error) {
```
kstola2: Remove this function, it will be on the other branch.

internal/contracts/contracts.go
```diff
@@ -274,4 +275,53 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
+	bKey, err := publickey.Encode(c.SenderPubKey)
+	if err != nil{
+		return bytes.Equal(bKey, pk.Bytes), errors.New("failed to encode sender public key")
+	}
+	// return the comparison of the above encoding with Bytes field of pk
+    return  bytes.Equal(bKey, pk.Bytes), nil
+	
+}
+
+func(c *Contract) GenerateRandomContract() (Contract){
```
kstola2: - [ ] I don't believe we need a receiver, remove it
- [ ] Remove parentheses from return type

internal/contracts/contracts.go
```diff
@@ -274,4 +275,53 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
+	
+	
+	genSignature := rand.Read(make([]byte, genSigLen));
+	
+	genValue := binary.LittleEndian.Uint64(b[:])
+	
+	genStateNonce := binary.LittleEndian.Uint64(b[:])
+	
+
+	c := &Contract{
```
kstola2: - [ ] instead of assigning `c` to this, just return it.

internal/contracts/contracts.go
```diff
@@ -274,4 +275,53 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
+		Version:  genVersion,
+		SenderPubKey: genSenderPubKey,
+		SigLen:  genSigLen,
+		Signature:  genSignature,
+		RecipPubKeyHash:  genrecipPubKeyHash,
+		Value:  genValue,
+		StateNonce:  genStateNonce,
+		
+	}
+	return c, nil
```
kstola2: - [ ] remove this return in favor of the above change.

internal/contracts/contracts.go
```diff
@@ -278,41 +278,20 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
-		return bytes.Equal(bKey, pk.Bytes), errors.New("failed to encode sender public key")
-	}
-	// return the comparison of the above encoding with Bytes field of pk
-    return  bytes.Equal(bKey, pk.Bytes), nil
-	
-}
 
-func(c *Contract) GenerateRandomContract() (Contract){
+
+funcGenerateRandomContract() Contract{
```
harshalpatel165: space

internal/contracts/contracts.go
```diff
@@ -274,4 +275,31 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
+	
+	genSenderPubKey := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	genrecipPubKeyHash := rand.Read(make([]byte, 32))
+	genVersion := binary.LittleEndian.Uint16(b[:])
+	genSigLen := binary.LittleEndian.Uint8(b[:]);
+	genSignature := rand.Read(make([]byte, genSigLen));
+	genValue := binary.LittleEndian.Uint64(b[:])
+	genStateNonce := binary.LittleEndian.Uint64(b[:])
+	
+	return &Contract{
```
harshalpatel165: will remove &

internal/contracts/contracts.go
```diff
@@ -274,4 +275,31 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
+		Version:  genVersion,
+		SenderPubKey: genSenderPubKey,
+		SigLen:  genSigLen,
+		Signature:  genSignature,
+		RecipPubKeyHash:  genrecipPubKeyHash,
+		Value:  genValue,
+		StateNonce:  genStateNonce,
+		
+	}, nil
```
kstola2: - [ ] There is no need to return a `nil` here, since your function only returns a `Contract` per the signature.

internal/contracts/contracts.go
```diff
@@ -274,4 +275,31 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
 		mc.StateNonce,
 	}
 	return c, nil
+	fmt.print(c)
```
kstola2: - [ ] remove print call here

internal/contracts/contracts.go
```diff
@@ -274,4 +275,31 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
 	return c, nil
+	
+}
+
+
+
+func GenerateRandomContract() Contract{
+	b := rand.Read(make([]byte, 32))
+	
+	genSenderPubKey := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```
kstola2: - [ ] `ecdsa.GenerateKey` generates a private key. You need to extract the public key from the private key

How this is done: https://github.com/SIGBlockchain/project_aurum/blob/master/internal/publickey/publickey_test.go#L56

internal/contracts/contracts.go
```diff
@@ -274,4 +275,31 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
+	return Contract{
+		Version:  genVersion,
+		SenderPubKey: genSenderPubKey,
+		SigLen:  genSigLen,
+		Signature:  genSignature,
+		RecipPubKeyHash:  genrecipPubKeyHash,
+		Value:  genValue,
+		StateNonce:  genStateNonce,
+		
+	},
```
kstola2: - [ ] Remove this comma

internal/contracts/contracts.go
```diff
@@ -274,4 +275,31 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
+	
+	genSenderPubKey := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	genrecipPubKeyHash := rand.Read(make([]byte, 32))
+	genVersion := binary.LittleEndian.Uint16(b[:])
+	genSigLen := binary.LittleEndian.Uint8(b[:]);
+	genSignature := rand.Read(make([]byte, genSigLen));
+	genValue := binary.LittleEndian.Uint64(b[:])
+	genStateNonce := binary.LittleEndian.Uint64(b[:])
+	
+	return Contract{
```
kstola2: Instead of defining the struct with `Contract { ... } `, leverage the `New` function instead that returns a contract, just feed it your `gen_` parameters.

internal/contracts/contracts.go
```diff
@@ -274,4 +275,22 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
+
+
+func GenerateRandomContract() *Contract{
+	b := rand.Read(make([]byte, 32))
+	
+	genSenderPrivKey := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	genrecipPubKeyHash := rand.Read(make([]byte, 32))
+	genVersion := binary.LittleEndian.Uint16(b[:])
+	genSigLen := binary.LittleEndian.Uint8(b[:]);
+	genSignature := rand.Read(make([]byte, genSigLen));
```
kstola2: Let's remove these two `Sig` variables, since we don't need them in the `New` function

internal/contracts/contracts.go
```diff
@@ -274,4 +275,22 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
+	
+}
+
+
+
+func GenerateRandomContract() *Contract{
+	b := rand.Read(make([]byte, 32))
+	
+	genSenderPrivKey := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	genrecipPubKeyHash := rand.Read(make([]byte, 32))
```
kstola2: - [ ] Fix camelCase format

internal/contracts/contracts.go
```diff
@@ -274,4 +276,20 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
+}
+
+
+
+func GenerateRandomContract() (*Contract,error){
+	b := make([]byte, 32)
+	rand.Read(b)
+	genSenderPrivKey, _ := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	genRecipPubKeyHash :=  b
+	genVersion := binary.LittleEndian.Uint16(b[:])
```
kstola2: Is there a way to make sure this is never 0? The only instance `New` returns an error is if the version passed in is a zero.

- [ ] Generate only nonzero numbers for version
- [ ] Make return type be just a `Contract`

harshalpatel165: The new function returns 2 values, contract and error, so when I return a new *random contract, it also returns error

kstola2: I see. We can do what you had initially: define a return variable, with its error, then just return the return variable at the end.  It doesn't make much sense to generate a random contract and need to do error checks on the call.

harshalpatel165: Also, the new function already checks if the version is greater than 0. do u want it so that is doesn't make a contract with a zero value version in the first place?

kstola2: Yes this is exactly what I want.

harshalpatel165:
I had to use error because rand.Read returns an error

internal/contracts/contracts.go
```diff
@@ -274,4 +276,34 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
 	}
 	return c, nil
+	
+}
+
+
+
+func GenerateRandomContract() (*Contract, error){
+	b := make([]byte, 32)
+	_, err :=rand.Read(b)
```
kstola2: When will this return a non-nil error

harshalpatel165: The read function from rand is a helper function that uses io.readfull function from. The readFull function reads the length of the bytes and returns an error if fewer bytes were read. So, if for some reason not all the bytes were read, it would return an error

kstola2: If there's no identifiable reason that all the bytes will be read, then ignore the error. I don't want to approve a generate random contract function that returns any errors because that is impractical.

harshalpatel165:
not sure why jsontest file was changed

internal/contracts/contracts.go
```diff
@@ -273,5 +275,30 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
+	
 }
+
+
+
+func GenerateRandomContract() (*Contract){
+	min := 1
+	maxVer := 65535
+	b := make([]byte, 32)
+	mrand.Read(b)
```
kstola2: What is the purpose of this?

internal/contracts/contracts.go
```diff
@@ -273,5 +275,30 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
...
+	genRecipPubKeyHash := b
```
kstola2: Can you make this the result of the hash of a random number? 

i.e. `hashing.New([]byte{ someRandomNumber })`

harshalpatel165: does it matter how long the number inside the byte is, as in does it matter if it is a random Uint16 or Uint64

harshalpatel165: I thought an array of bytes was multiple values, not just one singular number

kstola2: > does it matter how long the number inside the byte is, as in does it matter if it is a random Uint16 or Uint64

I don't think it matters. Choose whichever is easier for you

harshalpatel165: is it possible to have an array of bytes with just a single number

harshalpatel165: i thought it was an array with multiple numbers

internal/contracts/contracts.go
```diff
@@ -284,9 +284,10 @@ func (mc *JSONContract) Unmarshal() (Contract, error) {
 func GenerateRandomContract() (*Contract){
 	min := 1
 	maxVer := 65535
-	b := make([]byte, 32)
-	mrand.Read(b)
-	genRecipPubKeyHash := b 
+	b := make([]byte, 8)
+	binary.LittleEndian.PutUint64(b, mrand.Uint64())
+
+	genRecipPubKeyHash := hashing.New(b)
```
kstola2: `b` isn't random

harshalpatel165: I tested it out on playground, and it generates a random []byte every time I call the function

PR #430, opened 2019-11-04
completed issue410

bashbashbash:
Please provide help coming up with better comments for receiver functions

### DIFF
--- a/internal/accountstable/accountstable.go
+++ b/internal/accountstable/accountstable.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"encoding/hex"
 	"errors"
+	"sync"
 
 	"github.com/SIGBlockchain/project_aurum/internal/accountinfo"
 	"github.com/SIGBlockchain/project_aurum/internal/block"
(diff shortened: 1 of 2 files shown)


internal/accountstable/accountstable.go
```diff
@@ -14,6 +15,86 @@ import (
 	"github.com/SIGBlockchain/project_aurum/internal/sqlstatements"
 )
 
+// structs
+
+// ATConnection struct used for
+type ATConnection struct {
```
kstola2: Can we rename this to `AccountsTableConnection`?

internal/accountstable/accountstable.go
```diff
@@ -14,6 +15,86 @@ import (
...
+	lock   sync.RWMutex
+	dbconn *sql.DB
+}
+
+// Lock locks at for writing. If the lock is already locked for reading or writing,
```
kstola2: RW locks lock when writing, they never lock for reading

internal/accountstable/accountstable.go
```diff
@@ -14,6 +15,86 @@ import (
...
+	at.lock.Lock()
+}
+
+// Unlock unlocks at for writing. It is a run-time error if at is not locked for
+// writing on entry to Unlock.
+func (at *ATConnection) Unlock() {
+	at.lock.Unlock()
+}
+
+// InsertAccountIntoAccountBalanceTable is called from receiver in order to enforce interface
```
kstola2: "Calls accountstable.InsertAccountIntoAccountBalanaceTable using struct connection"

Same for all other TODO's; just replace the function name

internal/accountstable/accountstable.go
bashbashbash: Originally I was going to do this but the go-linter said this:

[click this link](https://ibb.co/CJGq515)

Thoughts?

kstola2: Interesting. I guess we can make it that.

In your comments can you replace `at` with `accounts table`, because right now it looks like the word "at" and it's a bit confusing.

Issue #432, opened 2019-11-09
AAA Test and Function GetBlockFromResponse


bashbashbash:
A AAA unit-testable response handler, which will convert the block in the response body from JSON to a normal block and return it

bashbashbash:
Completed

PR #433, opened 2019-11-09
Getblockfrombody AAA test and function

bashbashbash:
#432

### DIFF
--- a/internal/handlers/handlers.go
+++ b/internal/handlers/handlers.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"database/sql"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"strconv"
@@ -142,9 +143,12 @@ func HandleGetJSONBlockByHeight(fetcher ifaces.IBlockFetcher) func(w http.Respon
 // Note - per the documentation on the response struct, the body is never nil
 // TODO how do we test the response is valid if the body is never nil?
 func GetBlockFromResponse(r *http.Response) block.Block {
-	var requestBody block.Block
+	var requestBody block.JSONBlock
 	buf := new(bytes.Buffer)
 	buf.ReadFrom(r.Body)
 	json.Unmarshal(buf.Bytes(), &requestBody)
-	return requestBody
+
+	fmt.Printf("%v", requestBody)
+	bodyBlock, _ := requestBody.Unmarshal()
+	return bodyBlock
 }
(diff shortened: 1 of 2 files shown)


internal/handlers/handlers.go
```diff
@@ -137,3 +139,20 @@ func HandleGetJSONBlockByHeight(fetcher ifaces.IBlockFetcher) func(w http.Respon
 		io.WriteString(w, string(marshalledBlock))
 	}
 }
+
+// GetBlockFromResponse will convert the body of a reponse handler and return a Block
```
kstola2: converts the body of a response* , I don't know what a response handler is

internal/handlers/handlers.go
```diff
@@ -137,3 +139,20 @@ func HandleGetJSONBlockByHeight(fetcher ifaces.IBlockFetcher) func(w http.Respon
...
+// Note - per the documentation on the response struct, the body is never nil
+// TODO how do we test the response is valid if the body is never nil?
```
kstola2: HTTP uses status codes. Please read up on them. 200 would be returned upon successful block retrieval

- [ ] remove TODO

internal/handlers/handlers.go
```diff
@@ -137,3 +139,20 @@ func HandleGetJSONBlockByHeight(fetcher ifaces.IBlockFetcher) func(w http.Respon
...
+	if r.StatusCode == http.StatusBadRequest {
+		return block.Block{}, errors.New("StatusBadRequest")
+	}
+	var requestBody block.JSONBlock
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(r.Body)
+	json.Unmarshal(buf.Bytes(), &requestBody)
+
+	fmt.Printf("%v", requestBody)
+	bodyBlock, _ := requestBody.Unmarshal()
```
kstola2: - [ ] Handle the error returned from `Unmarshal`

internal/handlers/handlers_test.go
```diff
@@ -414,3 +414,71 @@ func TestGetJSONBlockByHeight(t *testing.T) {
...
+			rr := httptest.NewRecorder()
+
+			// Serve the request
+			handler.ServeHTTP(rr, req)
+
+			// Actual block that is recorded in the response
+			actualBlock, err := GetBlockFromResponse(rr.Result())
+
+			// Assert
+			if !reflect.DeepEqual(test.expectedBlock, actualBlock) {
```
kstola2: - [ ] Please use this function for comparisons: https://github.com/SIGBlockchain/project_aurum/blob/master/internal/block/block.go#L147

internal/handlers/handlers.go
bashbashbash: - [ ]

Issue #378, opened 2019-10-13
Make account info handler robust


kstola2:
- [ ] Have the handler for account info requests with a new wallet address check the pending pool first for the wallet address and return that information if it's found
- [ ] Test supports new behavior

bashbashbash:
I either need more explanation about this task or the functionality has already been implemented

PR #434, opened 2019-11-16
Functionized settings/main

bashbashbash:
#378 

The main goal was to simply settings/main, which I believe has been achieved. Pleaser provide comments.

Based on how I refactored settings/main, we do not need a test for that file. The functions that were placed into config.go don't have tests, but because they are straightforward, I am not sure they need tests. Please provide comments.

The test in validation is simple because I think it can be mocked, but first I wanted to check that mocking was appropriate, so I provided a simple test. Could make it more robust if desired. Please provide comments.


Vineet77:
There was a bug with parsing the version flag. I fixed it. Otherwise everything looks good.

